PR #8, opened 2018-08-15
Assorted edx-django-utils clean-up.

robrap:
**Description:** 

- Docs and README clean-up.
- Python3 support completed.
- Address review comments from edx-django-utils/pull/2.
- Add latest cookiecutter updates.

**JIRA:** 

ARCH-220

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed
- [x] PR author is listed in AUTHORS

**Post merge:**
- [ ] Create a tag
      finished.


robrap:
@ormsbee: I indicated all the file updates that were just coming from the cookiecutter so you could skip those.  If you need me to pull those out, let me know.  I can also delete those comments if they are a distraction.  Thanks.

Also includes fixes for your comments in https://github.com/edx/edx-django-utils/pull/2

edx_django_utils/cache/utils.py
```diff
@@ -335,25 +338,18 @@ def __nonzero__(self):
...
+    def __eq__(self, other):
+        if isinstance(other, CachedResponse):
+            if self.is_hit != other.is_hit:
+                return False
+            if self.key != other.key:
+                return False
+            if self.is_hit:
+                if self.value != other.value:
+                    return False
+            return True
```
ormsbee: I realize why you did it this way, and I think it's correct. But it seems pretty awkward to follow. Maybe it'd be clearer with something like:

```python
def __eq__(self, other):
    if not isinstance(other, CachedResponse):
        raise CachedResponseError()
    if self.is_hit != other.is_hit:
        return False

    if self.is_hit:
        return (self.key, self.value) == (other.key, other.value)
    else:
        return self.key == other.key  # cache misses have no value attribute
```

?

A bit of a nit, but it does seem a little weird to have an `__init__` that yields a object with different attributes set depending on values. I feel like that works great for languages that actually support algebraic types, but it causes these kinds of weird contortions in Python. If cache misses still had `None` for a value, this whole thing could have been:

```python
def __eq__(self, other):
    if not isinstance(other, CachedResponse):
        raise CachedResponseError()
    return (self.is_hit, self.key, self.value) == (self.is_hit, other.key, other.value)
```

Which I think is the most natural way to do it, and doesn't seem that risky in terms of foot-gun behavior.

edx_django_utils/private_utils.py
```diff
@@ -0,0 +1,57 @@
...
+                    break
+            except StopIteration:
+                raise AssertionError(
+                    'Required middleware that {} depends on are either missing or out of order. The problem was '
+                    'found when looking for {} from the complete expected list {}.'.format(
+                        concerned_object,
+                        expected_middleware,
+                        required_middleware,
+                    )
+                )
```
ormsbee: Would this work?

```python
def _check_middleware_dependencies(concerned_object, required_middleware):
    declared_middleware = getattr(settings, 'MIDDLEWARE', None)
    if declared_middleware is None:
        declared_middleware = settings.MIDDLEWARE_CLASSES  # Django 1.8 support

    # Filter out all the middleware except the ones we care about for ordering.
    matching_middleware = [mw for mw in declared_middleware if mw in required_middleware]
    if required_middleware != matching_middleware:
        raise AssertionError(
            "{} requires middleware order {} but matching middleware was {}".format(
                concerned_object, required_middleware, matching_middleware
            )
        )
```

edx_django_utils/private_utils.py
```diff
@@ -0,0 +1,57 @@
+"""
+Private utility methods.
+
+These utilities are designated for internal use only (i.e. inside
+edx-django-utils).  In some cases, these utilities could be elevated to public
+methods once they get some burn-in time and find a more permanent home in the
+library.
```
ormsbee: Nit: I think it's fine to make a module for middleware utilities and have no public methods in it (just _ prefixed ones), to indicate internal use, instead of having a global catch bucket for internal utilities.

robrap: I wanted to avoid ending up with a lot of directories each of which might have 1 helper function, but maybe it is fine.  If I set up a whole directory, do we want to make this public?  For example, the DRF Middleware I added could use this.

ormsbee: Actually, let's just leave it. It's probably not worth the length of this conversation thread. :P

robrap:
@ormsbee: Thanks for the cleaner Python.  Please add thumb when you are ready and I'll squash.  Thanks.

robrap:
You beat me by 17 seconds.

PR #17, opened 2018-12-14
Add user ids to New Relic as a custom parameter.

calistenson:
**Description:**

User ids as a custom parameter to New Relic data.

**JIRA:**

https://openedx.atlassian.net/browse/REVE-69

**Merge deadline:**

Today - ASAP if possible.

**Reviewers:**
- [ ] @edx/arch-review
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


edx_django_utils/monitoring/middleware.py
```diff
@@ -63,13 +63,15 @@ def accumulate_metric(cls, name, value):
...
     @classmethod
-    def _batch_report(cls):
+    def _batch_report(cls, request):
         """
         Report the collected custom metrics to New Relic.
         """
         if not newrelic:
             return
         metrics_cache = cls._get_metrics_cache()
+        if request.user.id:
```
mulby: I would also check request.user, since I think that will be None if the user isn't logged in... something like:

```python
if getattr(request, 'user') is not None and request.user.id:
    newrelic.agent.add_custom_parameter('user_id', request.user.id)
```

robrap:
@calistenson @mulby: Is this ok to do from a PII standpoint?  The [id is designated as PII](https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst) in the OEP.
Maybe have @macdiesel look at this?

douglashall:
I think you just need to bump the package version here: https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/__init__.py#L7

calistenson:
> @calistenson @mulby: Is this ok to do from a PII standpoint? The [id is designated as PII](https://github.com/edx/open-edx-proposals/blob/master/oeps/oep-0030-arch-pii-markup-and-auditing.rst) in the OEP.
> Maybe have @macdiesel look at this?

@robrap I discussed this with Brian Beggs a bit in Arch lunch and since New Relic dumps its data we should be okay from a GDPR standpoint. With PII in particular, since it's the user id and not the user name we thought we would be okay here. We'd like this information for the revenue launch on Monday but if you feel strongly enough we can pause.

edx_django_utils/monitoring/middleware.py
```diff
@@ -63,13 +63,17 @@ def accumulate_metric(cls, name, value):
...
-    def _batch_report(cls):
+    def _batch_report(cls, request):
         """
         Report the collected custom metrics to New Relic.
         """
         if not newrelic:
             return
         metrics_cache = cls._get_metrics_cache()
+        try:
+            newrelic.agent.add_custom_parameter('user_id', request.user.id)
```
nasthagiri: I see this is a rushed PR in trying to make a deadline.  So this comment is for a subsequent revision.

It seems odd to me that we are adding a _specific_ custom metric in a middleware that is providing _support_ for custom metrics - but not adding any specific ones itself.  We add specific custom metrics in other places in the code.  If we want to have a metric that we add to all of our APIs, for example, it should belong in [RequestMetricsMiddleware](https://github.com/edx/edx-drf-extensions/blob/0f69c806e85b84ad35efe0dd48cdf16619437972/edx_rest_framework_extensions/middleware.py#L7).  If my understanding of this separation of concerns is incorrect then we should update the docstring of this middleware.  Thanks.  :D

robrap: @nasthagiri: What you say is true.  I must admit, I didn't really review the code - even though I merged.  I was counting on the other set of approvals.  Sorry.
@calistenson: Is this a refactor you can do?

calistenson: @nasthagiri, @robrap thanks for this note. I jumped to changing this repo after direction from Dave and did not know that the RequestMetricsMiddleware repo existed. I'm sure that a refactor is possible, but I'd prefer to check in with @mulby and team and then figure out how to proceed from there.

nasthagiri: To be clear - a refactor and follow-on can happen after the weekend.  Let's take this as an "IOU" to clean it up - as we don't want to necessarily block the team's immediate deadline.  Hopefully, it's just a matter of moving this metric to the other middleware - and not much more.

PR #26, opened 2019-07-10
Support Django 1.10 style middleware

mikix:
In Django 1.10, a new middleware style was introduced, that uses a MIDDLEWARE setting rather than MIDDLEWARE_CLASSES.

In Django 2, only the new style will be supported.

This change allows our middleware to be used both ways, but drops support for Django 1.8.

This package claims to support Django 2, but the current middleware classes can't work with Django 2, right? Was that ever tried?

If Django 1.8 support is important to maintain, I can do an optional import situation and keep support. But I figured the open edx ecosystem is on 1.11+.

**Reviewers:**
- [ ] @edx/arch-review
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/private_utils.py
+++ b/edx_django_utils/private_utils.py
@@ -31,12 +31,12 @@ def _check_middleware_dependencies(concerned_object, required_middleware):
 
     Raises:
         AssertionError if the provided dependencies don't appear in
-            MIDDLEWARE_CLASSES in the correct order.
+            MIDDLEWARE in the correct order.
 
     """
     declared_middleware = getattr(settings, 'MIDDLEWARE', None)
     if declared_middleware is None:
-        declared_middleware = settings.MIDDLEWARE_CLASSES  # Django 1.8 support
+        declared_middleware = settings.MIDDLEWARE_CLASSES  # Pre-Django 2 support
 
     # Filter out all the middleware except the ones we care about for ordering.
     matching_middleware = [mw for mw in declared_middleware if mw in required_middleware]
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ python:
   - 3.6
 
 env:
-  - TOXENV=django18
   - TOXENV=django111
   - TOXENV=django20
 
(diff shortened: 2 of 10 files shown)


edx_django_utils/private_utils.py
```diff
@@ -36,7 +36,7 @@ def __init__(self):
     """
     declared_middleware = getattr(settings, 'MIDDLEWARE', None)
     if declared_middleware is None:
-        declared_middleware = settings.MIDDLEWARE_CLASSES  # Django 1.8 support
+        declared_middleware = settings.MIDDLEWARE_CLASSES  # Pre-Django 2 support
```
robrap: Does this mean it is backward compatible?  If so, why bump the version to `2.0.0`?

mikix: It is backwards compatible through 1.10. This code was previously trying to be forward compatible, but I don't think it ever was.

I can update the docs to use new-style only, sure. Both are fine, and old-style will only go away once we stop supporting django 1.x. But since we are all on 1.11, new style is recommended path, so docs should push it.

mikix: Oh, and I bumped the version to `2.0.0` because I'm dropping support for django 1.8

mikix:
I haven't tested yet with a new-style repo. I'm trying to do that with course-discovery now.

But the change *should* be harmless. We're doing it elsewhere: https://github.com/edx/edx-drf-extensions/commit/84cf1de98338d6a5f11e26158481c2158005151c

But will try to set up a course-discovery PR that points to this branch.

mikix:
https://github.com/edx/course-discovery/pull/2011

^ That seems to work with this branch. And I also manually tested that still using MIDDLEWARE_CLASSES with this branch continues to work.

mikix:
Oh I did update the docs in this PR

robrap:
Oops.  Not sure how I missed that.  Thanks.

PR #31, opened 2020-02-12
Removed python 2.7 support

davestgermain:


.travis.yml
```diff
@@ -3,7 +3,6 @@
 language: python
 
 python:
-  - 2.7
```
robrap: Wondering about timing of this change given this library is used everywhere, including Insights, which is not yet on Python 3.  May be fine to constrain in Insights, as long as we don't run into any security issues, etc.
@jmbowman: Any thoughts on whether it is better to wait or it is fine?

edx_django_utils/cache/tests/test_middleware.py
```diff
@@ -2,8 +2,6 @@
 """
 Tests for the RequestCacheMiddleware.
 """
-# pylint: disable=missing-docstring
```
robrap: Maybe this is just personal preferences, but I added the disable pylint to avoid docstrings that basically just repeat the names of the classes.  So, I personally prefer not making these changes.

requirements/constraints.txt
```diff
@@ -10,3 +10,6 @@
 
  # These packages are backports which can only be installed on Python 2.7
 futures ; python_version == "2.7"
+Django<3.0
+mock<4.0  # requires python>3.5
+zipp<2.0
```
robrap: Please add a comment for all constraints to explain why they are needed, which will make it simpler to know when they can go away.  Thanks.

.travis.yml
jmbowman: The last time this was discussed, we decided it was ok to drop Python 2.7 support for libraries used in Insights, since it can just avoid upgrading them.  If for some reason we do need Insights to get a fixed version of the package before it finishes the Python 3 upgrade, we can branch from the commit before this merges and do a new 2.x release.

robrap: 👍

davestgermain:
@robrap I'm not allowed to merge this. Could you?

edx_django_utils/cache/tests/test_middleware.py
davestgermain: Ok, I reverted those redundant docstrings. I had thought that pylint was complaining about it, somehow.

PR #33, opened 2020-03-19
Add tooling for working with the read-replica database.

cpennington:
This migrates the use_read_replica_if_available and read_replica_or_default
functions from edx_platform/common/djangoapps/utils/query.py, and augments
them with a pair of context managers for asserting that particular chunks
of code are read-only (and thus read-replica safe).

**Description:**

Describe in a couple of sentence what this PR adds

**JIRA:**

[XXX-XXXX](https://openedx.atlassian.net/browse/XXX-XXXX)

**Dependencies:**

List dependencies on other outstanding PRs, issues, etc.

**Merge deadline:**

List merge deadline (if any)

List any non-trivial installation
instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happened instead - check failed.

**Reviewers:**
- [ ] @edx/arch-review (Required)
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

migrations, etc.

### DIFF
--- a/edx_django_utils/db/read_replica.py
+++ b/edx_django_utils/db/read_replica.py
@@ -1,0 +1,157 @@
+"""
+Tools for making queries read from the read-replica database, rather than from the writer database.
+
+The read-replica can be used for
+    a. long-running queries that aren't time sensitive
+    b. reads of rows that are frequently written, but where reads can be eventually consistent
+
+Settings:
+    EDX_READ_REPLICA_DB_NAME: The name of the read-replica in the DATABASES django setting.
+        Defaults to "read_replica".
+    EDX_WRITER_DB_NAME: The name of the writer in the DATABASES django setting.
+        Defaults to "default".
+"""
+import threading
+from contextlib import contextmanager
+
+from django.conf import settings
+
+READ_REPLICA_NAME = getattr(settings, "EDX_READ_REPLICA_DB_NAME", "read_replica")
+WRITER_NAME = getattr(settings, "EDX_WRITER_DB_NAME", "default")
+
+READ_REPLICA_OR_DEFAULT = (
+    READ_REPLICA_NAME if READ_REPLICA_NAME in settings.DATABASES else WRITER_NAME
+)
+
+
+def use_read_replica_if_available(queryset):
+    """
+    If there is a database called 'read_replica',
+    use that database for the queryset / manager.
+
+    Example usage:
+        queryset = use_read_replica_if_available(SomeModel.objects.filter(...))
+
+    Arguments:
+        queryset (QuerySet)
+
+    Returns: QuerySet
+    """
+    return queryset.using(READ_REPLICA_OR_DEFAULT)
+
+
+def read_replica_or_default():
+    """
+    If there is a database called READ_REPLICA_DB,
+    return READ_REPLICA_DB, otherwise return WRITER_NAME.
+
+    This function is similiar to `use_read_replica_if_available`,
+    but is be more syntactically convenient for method call chaining.
+    Also, it always falls back to WRITER_NAME,
+    no matter what the queryset was using before.
+
+    Example usage:
+        queryset = SomeModel.objects.filter(...).using(read_replica_or_default())
+
+    Returns: str
+    """
+    return READ_REPLICA_OR_DEFAULT
+
+
+@contextmanager
+def read_queries_only():
+    """
+    A context manager that sets all reads inside it to be from the read-replica.
+
+    It is an error to call this from inside a write_queries context.
+
+    The ReadReplicaRouter must be used for this decorator to affect queries.
+    """
+    old_db_name = _storage.db_name
+    assert (
+        old_db_name is None or old_db_name == READ_REPLICA_NAME
+    ), "Can't use read_queries_only inside a write_queries contextmanager"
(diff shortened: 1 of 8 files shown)


edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,134 @@
+"""
+Tools for making queries read from the read-replica database, rather than from the writer database.
```
robrap: I personally would find it helpful if you could provide some guidance on when these should be used.  Should you strive to use the read-replica for any querying that can use it?  Is it more important to get long running queries to use it?  Other?

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,134 @@
...
+
+    Returns: str
+    """
+    return READ_REPLICA_OR_DEFAULT
+
+
+@contextmanager
+def read_queries_only(db_name):
+    """
+    A context manager that sets all reads inside it to be from the read-replica.
```
robrap: It seems like this won't fall back to default.  What would happen?  Can you add more comments around this?  Also, maybe rename to `read_replica_queries_only` if this is the case?

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,134 @@
...
+    try:
+        yield
+    finally:
+        _storage.db_name = old_db_name
+
+
+@contextmanager
+def write_queries(db_name):
+    """
+    A context manager that sets all reads inside it to be from the writer.
```
robrap: The name might be fine.  Could you provide more context on when someone would want to use this?  That may indicate a name change, but maybe not.

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+_storage = _ReadReplicaRouterStorage()
+
+
+class ReadReplicaRouter:
+    """
+    A database router that by default, reads from the writer database,
+    but can be overridden with a context manager to route all reads
+    to the read-replica.
```
robrap: Would you mind adding the following?  I wasn't clear where this API was coming from.
```
See https://docs.djangoproject.com/en/2.2/topics/db/multi-db/#automatic-database-routing
```

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,137 @@
...
+
+from django.conf import settings
+
+
+READ_REPLICA_NAME = getattr(settings, "EDX_READ_REPLICA_DB_NAME", "read_replica")
+WRITER_NAME = getattr(settings, "EDX_WRITER_DB_NAME", "default")
+
+READ_REPLICA_OR_DEFAULT = (
+    READ_REPLICA_NAME if READ_REPLICA_NAME in settings.DATABASES else WRITER_NAME
+)
```
ormsbee: It's safe to do this because Django settings are lazy, right?

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,158 @@
...
+class _ReadReplicaRouterStorage(threading.local):
+    def __init__(self):
+        super(_ReadReplicaRouterStorage, self).__init__()
+        self.db_name = None
+
+
+_storage = _ReadReplicaRouterStorage()
+
+
+class ReadReplicaRouter:
```
ormsbee: This doesn't need to be a thread local?

edx_django_utils/db/read_replica.py
```diff
@@ -0,0 +1,158 @@
...
+
+The read-replica can be used for
+    a. long-running queries that aren't time sensitive
+    b. reads of rows that are frequently written, but where reads can be eventually consistent
+
+Settings:
+    EDX_READ_REPLICA_DB_NAME: The name of the read-replica in the DATABASES django setting.
+        Defaults to "read_replica".
+    EDX_WRITER_DB_NAME: The name of the writer in the DATABASES django setting.
+        Defaults to "writer".
```
feanil: s/writer/default/

edx_django_utils/db/read_replica.py
cpennington: The router isn't threadlocal. The storage is.

edx_django_utils/db/read_replica.py
cpennington: Fixed

edx_django_utils/db/read_replica.py
cpennington: Are you thinking about the fact that the database router is going to be referencing this file? The router will be named via a string, so it should be ok (it'll be loaded later)

ormsbee: I was mostly thinking that I didn't know if Django's settings were always properly initialized here because these imports run at import, but I think that was only a problem on much earlier versions of Django (but didn't remember for sure). I think it's fine these days, but I just wanted to double check.

cpennington:
The build is passing, but not reporting to github (https://travis-ci.org/github/edx/edx-django-utils/builds/666356074). I'm going to merge this.

PR #45, opened 2020-07-21
ARCHBOM-1357: add catch-all to CodeOwnerMetricMiddleware

robrap:
**Description:**

Added a catch-all capability to CodeOwnerMetricMiddleware
when CODE_OWNER_MAPPINGS includes a '*' as a team's
module. The catch-all is used only if there is no other
match.

**JIRA:**

[ARCHBOM-1357](https://openedx.atlassian.net/browse/ARCHBOM-1357)

**Reviewers:**
- [x] @edx/arch-review (Required)

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [X] Version bumped
- [X] Changelog record added
- [X] Documentation updated (not only docstrings)
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/monitoring/code_owner/middleware.py
+++ b/edx_django_utils/monitoring/code_owner/middleware.py
@@ -16,6 +16,9 @@ class CodeOwnerMetricMiddleware:
     """
     Django middleware object to set custom metrics for the owner of each view.
 
+    Uses Django Setting CODE_OWNER_MAPPINGS as documented in get_code_owner_from_module.
+    Note: Use '*' as a configured module if you want a catch-all when there is no other match.
+
     Custom metrics set:
     - code_owner: The owning team mapped to the current view.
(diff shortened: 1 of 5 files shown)


edx_django_utils/monitoring/code_owner/middleware.py
```diff
@@ -51,6 +54,9 @@ def _set_code_owner_metric(self, request):
             return
         if not path_error:
             # module found, but mapping wasn't configured
+            code_owner = self._set_code_owner_metric_catch_all()
+            if code_owner:
+                set_custom_metric('code_owner', code_owner)
```
robrap, line 59: **Reviewer Note**: Although this code would be less noisy if `_set_code_owner_metric_catch_all()` set the `code_owner` metric, I went with this option so that this method wouldn't be setting `code_owner` for some cases (the existing code), and not for other cases (the new code).

PR #58, opened 2020-08-29
ARCHBOM-1490: fix custom monitoring accumulate

robrap:
**Description:**

The accumulate method wasn't accumulating, and the unit
test wasn't working as expected.

**JIRA:**

[ARCHBOM-1490](https://openedx.atlassian.net/browse/ARCHBOM-1490)

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/monitoring/middleware.py
+++ b/edx_django_utils/monitoring/middleware.py
@@ -67,7 +67,9 @@ class MonitoringCustomMetricsMiddleware(MiddlewareMixin):
             except TypeError:
                 _set_custom_attribute(
                     'error_adding_accumulated_metric',
-                    'new_value={}, cached_value={}'.format(repr(value), repr(cached_response.value))
+                    'name={}, new_value={}, cached_value={}'.format(
+                        name, repr(value), repr(cached_response.value)
+                    )
                 )
                 return
         else:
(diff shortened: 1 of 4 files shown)


edx_django_utils/monitoring/middleware.py
```diff
@@ -60,8 +60,19 @@ def accumulate_metric(cls, name, value):
...
-        metrics_cache.setdefault(name, 0)
-        metrics_cache.set(name, value)
+        cached_response = metrics_cache.get_cached_response(name)
+        if cached_response.is_found:
+            try:
+                accumulated_value = value + cached_response.value
+            except TypeError:
+                _set_custom_attribute(
+                    'error_adding_accumulated_metric',
+                    'new_value={}, cached_value={}'.format(value, cached_response.value)
```
jmbowman, line 70: If the main point of this message is to describe type errors, maybe the arguments should be wrapped in `repr()`?  Otherwise ints and strings could look the same, etc.

robrap: Added `repr`.

robrap: @jmbowman: You also made me realize that I forgot to add the `name`, so that has been added as well.

jmbowman: I don't see `name` added yet, have you not pushed that commit?

robrap: Good call.  Just pushed the change.

timmc-edx (approved):

PR #68, opened 2020-10-20
ARCHBOM-1527: update code owner mapping

robrap:
**Description:**

* Exposed existing get_code_owner_from_module via the public api.
* Fixed get_code_owner_from_module to not require a call to is_code_owner_mappings_configured beforehand.
* Set the existing code_owner_path_module custom attribute, even for cases where the transaction name was used, rather than the view module.
* Refactored and simplified configuration and tests.

**JIRA:**

[ARCHBOM-1527](https://openedx.atlassian.net/browse/ARCHBOM-1527)

**Merge checklist:**
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/monitoring/code_owner/tests/test_middleware.py
+++ b/edx_django_utils/monitoring/code_owner/tests/test_middleware.py
@@ -12,7 +12,7 @@ from mock import Mock, call, patch
 
 from edx_django_utils.monitoring.code_owner.middleware import CodeOwnerMonitoringMiddleware
 from edx_django_utils.monitoring.code_owner.tests.mock_views import MockViewTest
-from edx_django_utils.monitoring.code_owner.utils import _process_code_owner_mappings
+from edx_django_utils.monitoring.code_owner.utils import clear_cached_mappings
 
 
 class MockMiddlewareViewTest(View):
@@ -34,6 +34,7 @@ class CodeOwnerMetricMiddlewareTests(TestCase):
 
     def setUp(self):
         super().setUp()
+        clear_cached_mappings()
         self.mock_get_response = Mock()
         self.middleware = CodeOwnerMonitoringMiddleware(self.mock_get_response)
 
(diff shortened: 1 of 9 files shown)


edx_django_utils/monitoring/code_owner/tests/test_middleware.py
```diff
@@ -12,7 +12,7 @@
 
 from edx_django_utils.monitoring.code_owner.middleware import CodeOwnerMonitoringMiddleware
 from edx_django_utils.monitoring.code_owner.tests.mock_views import MockViewTest
-from edx_django_utils.monitoring.code_owner.utils import _process_code_owner_mappings
+from edx_django_utils.monitoring.code_owner.utils import clear_cached_mappings
```
robrap, line 15: **Reviewer Note**: Part of this PR simply provides a cleaner implementation of the mapping setup and testing. `clear_cached_mappings` is used in place of all the messy patches of `_PATH_TO_CODE_OWNER_MAPPINGS`.

edx_django_utils/monitoring/code_owner/tests/test_middleware.py
```diff
@@ -114,29 +107,35 @@ def test_code_owner_path_mapping_with_catch_all(
...
+        (
+            'edx_django_utils.monitoring.code_owner.tests.mock_views',
+            'edx_django_utils.monitoring.code_owner.tests.mock_views:MockViewTest',
+            'team-red'
+        ),
     )
     @ddt.unpack
     def test_code_owner_transaction_mapping_hits_and_misses(
-        self, transaction_name, expected_owner, mock_newrelic_agent, mock_set_custom_attribute
+        self, path_module, transaction_name, expected_owner, mock_newrelic_agent, mock_set_custom_attribute
```
robrap, line 123: **Reviewer Note**: This is a new feature when we provide the path in a custom attribute even for these cases where we need to use the transaction name to determine the code_owner.

edx_django_utils/monitoring/code_owner/tests/test_middleware.py
```diff
@@ -195,46 +195,33 @@ def test_code_owner_no_mappings(self, mock_set_custom_attribute):
...
-            expected_path_module = self._REQUEST_PATH_TO_MODULE_PATH['/test/']
-            self._assert_code_owner_custom_attributes(
-                mock_set_custom_attribute, path_module=expected_path_module,
-                has_path_error=True, has_transaction_error=True
-            )
+        request = RequestFactory().get('/test/')
+        self.middleware(request)
+        expected_path_module = self._REQUEST_PATH_TO_MODULE_PATH['/test/']
+        self._assert_code_owner_custom_attributes(
+            mock_set_custom_attribute, path_module=expected_path_module,
```
robrap, line 223: **Reviewer Note**: Don't worry about the fact that we are no longer providing errors in the custom attributes for this very unexpected case. This case will log an exception, which is tested elsewhere. Also, we would already be alerted from New Relic, providing we look for transactions with a path module and no code_owner, which we plan to do.

edx_django_utils/monitoring/code_owner/utils.py
```diff
@@ -61,48 +67,47 @@ def _process_code_owner_mappings():
         }
 
     """
-    _CODE_OWNER_MAPPINGS = getattr(settings, 'CODE_OWNER_MAPPINGS', None)
-    if not _CODE_OWNER_MAPPINGS:
+    global _PATH_TO_CODE_OWNER_MAPPINGS
+
+    # Return cached processed mappings if already processed
+    if _PATH_TO_CODE_OWNER_MAPPINGS is not None:
+        return _PATH_TO_CODE_OWNER_MAPPINGS
```
timmc-edx, line 74: I think there are thread-safety issues here that could be exposed on startup, as the first few requests race to complete this mapping. Maybe put a mutex around the method?

timmc-edx: (Everything else looks good.)

robrap: Silly me.  I looked up how to do this in Django, and not Python, and didn't see anything simple so I did a hacked solution.  I can update with a mutex to replace the need for a comment.

timmc-edx: That looks good! There's one other case that occurs to me, which is that the global could be cleared after the conditional check here but before the return, so that a couple requests return None instead of a dict. But that clearing should only happen during testing, so I'm not worried. :-)

timmc-edx (approved):

edx_django_utils/monitoring/code_owner/utils.py
robrap: Ok.  Leaving as-is.  Fast and cheap and won't lock.  Thanks.

PR #73, opened 2020-11-13
ARCHBOM-1260: enable code_owner for celery tasks

robrap:
**Description:**

* add set_code_owner_attribute decorator for use with celery tasks.
* clean up code owner middleware code.
* rename custom attribute code_owner_path_module to
code_owner_module. This may affect monitoring dashboards.
* minor change in when error custom attributes are set.

**JIRA:**

[ARCHBOM-1260](https://openedx.atlassian.net/browse/ARCHBOM-1260)

**Testing:**

WIP testing in edx-platform https://github.com/edx/edx-platform/pull/25584 (currently failing)

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/monitoring/docs/how_tos/add_code_owner_custom_attribute_to_an_ida.rst
+++ b/edx_django_utils/monitoring/docs/how_tos/add_code_owner_custom_attribute_to_an_ida.rst
@@ -21,6 +21,25 @@ Setting up the Middleware
 
 You simply need to add ``edx_django_utils.monitoring.CodeOwnerMonitoringMiddleware`` as described in the README to make this functionality available. Then it is ready to be configured.
 
+Handling celery tasks
+---------------------
+
+Celery tasks require use of a special decorator to set the ``code_owner`` custom attribute because no middleware will be run.
+
+Here is an example::
+
+  @task()
+  @set_code_owner_attribute
+  def example_task():
+      ...
+
+If the task is not compatible with additional decorators, you can use the following alternative::
+
+  @task()
+  def example_task():
+      set_code_owner_attribute_from_module(__name__)
+      ...
+
 Configuring your app settings
 -----------------------------
 
--- a/edx_django_utils/monitoring/internal/code_owner/middleware.py
+++ b/edx_django_utils/monitoring/internal/code_owner/middleware.py
@@ -4,10 +4,11 @@ Middleware for code_owner custom attribute
 import logging
 
 from django.urls import resolve
+from django.urls.exceptions import Resolver404
 
 from ..transactions import get_current_transaction
 from ..utils import set_custom_attribute
-from .utils import get_code_owner_from_module, is_code_owner_mappings_configured
+from .utils import _get_catch_all_code_owner, get_code_owner_from_module, is_code_owner_mappings_configured
 
 log = logging.getLogger(__name__)
 
@@ -21,10 +22,8 @@ class CodeOwnerMonitoringMiddleware:
 
     Custom attributes set:
     - code_owner: The owning team mapped to the current view.
-    - code_owner_mapping_error: If there are any errors when trying to perform the mapping.
+    - code_owner_module: The module found from the request or current transaction.
     - code_owner_path_error: The error mapping by path, if code_owner isn't found in other ways.
-    - code_owner_path_module: The __module__ of the view_func which was used to try to map to code_owner.
-        This can be used to find missing mappings.
     - code_owner_transaction_error: The error mapping by transaction, if code_owner isn't found in other ways.
     - code_owner_transaction_name: The current transaction name used to try to map to code_owner.
         This can be used to find missing mappings.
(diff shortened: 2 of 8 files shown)


edx_django_utils/monitoring/internal/code_owner/middleware.py
```diff
@@ -21,10 +21,8 @@ class CodeOwnerMonitoringMiddleware:
 
     Custom attributes set:
     - code_owner: The owning team mapped to the current view.
-    - code_owner_mapping_error: If there are any errors when trying to perform the mapping.
```
robrap, line 24: **Reviewer Note**: This simply didn't exist any more, so this was just clean up.

edx_django_utils/monitoring/internal/code_owner/middleware.py
```diff
@@ -21,10 +21,8 @@ class CodeOwnerMonitoringMiddleware:
...
+    - code_owner_module: The module found from the request or current transaction.
```
robrap, line 24: **Reviewer Note**: Simplified the name to make it more accurate, since it is used in multiple situations.

edx_django_utils/monitoring/internal/code_owner/middleware.py
```diff
@@ -43,105 +41,85 @@ def process_exception(self, request, exception):
...
+        Sets the code_owner custom attribute for the request.
         """
-        code_owner, path_error = self._set_code_owner_attribute_from_path(request)
-        if code_owner:
-            set_custom_attribute('code_owner', code_owner)
-            return
-        if not path_error:
-            # module found, but mapping wasn't configured
-            code_owner = self._set_code_owner_attribute_catch_all()
+        module = self._get_module_from_request(request)
```
robrap, line 46: @idegtiarov: This is some of the clean-up I had promised. Hopefully getting the module before getting the code_owner makes this code easier to understand.

edx_django_utils/monitoring/internal/code_owner/middleware.py
```diff
@@ -43,105 +41,85 @@ def process_exception(self, request, exception):
...
+        if module:
+            set_custom_attribute('code_owner_module', module)
+            return module
 
-        # only report errors if code_owner couldn't be found, including catch-all
+        # monitor errors if module was not found
         if path_error:
             set_custom_attribute('code_owner_path_error', path_error)
         if transaction_error:
             set_custom_attribute('code_owner_transaction_error', transaction_error)
```
robrap, line 87: **Reviewer Note**: This refactor ended up with a slight change in logic, where these error attributes are created even when we have a catch-all defined for code_owner. It's probably better this way if you actually want to find issues that resulted in the catch-all. :)

edx_django_utils/monitoring/internal/code_owner/utils.py
```diff
@@ -103,6 +109,60 @@ def get_code_owner_mappings():
...
+        @set_code_owner_attribute(__name__)
+        @task()
+        def example_task():
+            ...
+
+    """
+    def actual_decorator(wrapped_function):
+        @wraps(wrapped_function)
+        def new_function(*args, **kwargs):
+            _set_code_owner_attribute_from_module(module)
```
robrap, line 160: **Reviewer Note**: Originally I was using `wrapped_function.__module__`, rather than a module argument, but I feared that cases with compound decorators might report the wrong module.

robrap: 1. Some of the edx-platform tests are failing: https://github.com/edx/edx-platform/pull/25584. I thought this was right, but maybe I'm missing something.
2. Since @task needs to be outer-most, and it relies on the function name,maybe I can trust that `wrapped_function.__module__` would work?  Would removing the function call fix the failures?

robrap: I tried without the argument and I am getting the same error. I need to start debugging and/or get a second pair of eyes.

edx_django_utils/monitoring/internal/code_owner/utils.py
```diff
@@ -103,6 +109,58 @@ def get_code_owner_mappings():
...
+    If the catch-all module "*" is configured, return the code_owner.
+
+    Returns:
+        (str): code_owner or None if no catch-all configured.
+
+    """
+    try:
+        code_owner = get_code_owner_from_module('*')
+        return code_owner
+    except Exception:  # pylint: disable=broad-except; #pragma: no cover
```
regisb, line 123: I think we should really avoid this pattern. At the very least, we should be logging an exception:

```
...
except Exception as e:
    logger.exception(e)
    return None
```

But we should really be more specific about the type of exception that we are catching. In my experience, the only times where we need to catch Exception is when we want to convert to another type of exception. E.g:

```
except Exception as e:
    raise CodeOwnerException from e
```

edx_django_utils/monitoring/internal/code_owner/utils.py
```diff
@@ -103,6 +109,58 @@ def get_code_owner_mappings():
...
+    """
+    set_custom_attribute('code_owner_module', module)
+    code_owner = get_code_owner_from_module(module)
+    if code_owner:
+        set_custom_attribute('code_owner', code_owner)
+        return
+
+    code_owner = _get_catch_all_code_owner()
+    if code_owner:
+        set_custom_attribute('code_owner', code_owner)
```
regisb, line 139: This block can be simplified:

```
code_owner = get_code_owner_from_module(module)
if not code_owner:
    code_owner = _get_catch_all_code_owner()
set_custom_attribute('code_owner', code_owner)
```

robrap: It would still need a final `if code_owner` check, but makes sense.

edx_django_utils/monitoring/internal/code_owner/utils.py
robrap: This topic was already discussed on the last PR. I do not plan on making changes at this time. I don’t want to add risk to what I am trying to get done, and this cleanup could be handled separately.

edx_django_utils/monitoring/internal/code_owner/middleware.py
idegtiarov: Good to know! Thanks.

edx_django_utils/monitoring/internal/code_owner/utils.py
robrap: FYI: I added temporary custom attributes to all of the broad exceptions.  I imagine that these will simply be deleted, but I am being conservative. I will follow-up in another pass later.

edx_django_utils/monitoring/internal/code_owner/utils.py
robrap: FYI: I refactored the two places with this code.  It didn't shrink lines of code in one case, but it did remove the duplication around setting the custom attribute.

jinder1s (approved):

edx_django_utils/monitoring/docs/how_tos/add_code_owner_custom_attribute_to_an_ida.rst
```diff
@@ -21,6 +21,25 @@ Setting up the Middleware
...
+  def example_task():
+      ...
+
+If the task is not compatible with additional decorators, you can use the following alternative::
+
+  @task()
+  def example_task():
+      set_code_owner_attribute_from_module(__name__)
+      ...
+
```
robrap, line 42: @jinder1s: FYI: I made this minor update.

edx_django_utils/monitoring/internal/code_owner/utils.py
robrap: FYI: I left this as the cleaner alternative, since we now offer an alternative to using the decorator.

PR #76, opened 2020-11-17
add record_exception and update deprecated custom attribute values

robrap:
**Description:**

* Add `record_exception` to monitor uncaught exceptions.
* Add additional details to the `deprecated_monitoring_utils`
custom attribute values to make it simpler to track down usage.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_django_utils/monitoring/internal/utils.py
+++ b/edx_django_utils/monitoring/internal/utils.py
@@ -21,7 +21,7 @@ from .middleware import CachedCustomMonitoringMiddleware
 
 try:
     import newrelic.agent
-except ImportError:
+except ImportError:  # pragma: no cover
     newrelic = None  # pylint: disable=invalid-name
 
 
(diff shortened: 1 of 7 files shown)


edx_django_utils/monitoring/internal/utils.py
```diff
@@ -80,3 +80,11 @@ def set_custom_attribute(key, value):
     if newrelic:  # pragma: no cover
         # note: parameter is new relic's older name for attributes
         newrelic.agent.add_custom_parameter(key, value)
+
+
+def record_exception():
+    """
+    Records a caught exception to the monitoring system.
```
feanil, line 87:
```suggestion
    Records a caught exception to the monitoring system. By default only unhandled exceptions
    are caught.  This function can be called to record exceptions as errors even if you handle the
    exception gracefully from a user perspective.
    
    See https://docs.newrelic.com/docs/agents/python-agent/python-agent-api/recordexception-python-agent-api for more details.
```

feanil (approved):
Suggested some more help text to make the use of the function more clear.  You can take or replace as long as the intent is captured and then I don't think it needs another review.

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -80,3 +80,19 @@ def set_custom_attribute(key, value):
...
+
+    Note: By default, only unhandled exceptions are monitored. This function
+    can be called to record exceptions as monitored errors, even if you handle
+    the exception gracefully from a user perspective.
+
+    For more details, see:
+    https://docs.newrelic.com/docs/agents/python-agent/python-agent-api/recordexception-python-agent-api
```
robrap, line 94: @feanil: Thank you.  FYI: I adjusted the language a little bit.

PR #102, opened 2021-03-02
ENT-4236 | added chunked_queryset utility.

HammadAhmadWaqas:
**Description:**
      
      Added utity to Slice a queryset into chunks.
      
      The function slices a queryset into smaller QuerySets containing chunk_size objects and then yields them. It is
      used to avoid memory error when processing huge querysets, and also to avoid database errors due to the
      database pulling the whole table at once. Additionally, without using a chunked queryset, concurrent database
      modification while processing a large table might repeat or skip some entries.

**JIRA:**

[ENT-4236](https://openedx.atlassian.net/browse/ENT-4236)

**Dependencies:**

List dependencies on other outstanding PRs, issues, etc.

**Merge deadline:**

List merge deadline (if any)

List any non-trivial installation
instructions.

**Testing instructions:**

1. Open page A
2. Do thing B
3. Expect C to happen
4. If D happened instead - check failed.

**Reviewers:**
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

migrations, etc.


edx_django_utils/__init__.py
```diff
@@ -2,7 +2,7 @@
 EdX utilities for Django Application development..
 """
 
-__version__ = "3.14.0"
+__version__ = "3.14.1"
```
robrap, line 5: FYI: See https://open-edx-proposals.readthedocs.io/en/latest/oep-0047-bp-semantic-versioning.html to understand a better version update.

edx_django_utils/db/queryset_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+def chunked_queryset(queryset, chunk_size=100):
+    """
+    Slice a queryset into chunks.
+
+    This utility method is helpful for large QuerySets. As it is not recommended to load
+    whole QuerySet data at a same time. Using this method we can iterate over the large QuerySets
+    and load only a chunked piece of the QuerySets at a time.
+
+    Source: https://www.djangosnippets.org/snippets/10599/
```
robrap, line 14: I'd like @ormsbee review this and/or to point to docs he may have written on proper paging.

Here are the docs from this source:
> The function slices a queryset into smaller querysets containing chunk_size objects and then yield them. It is used to avoid memory error when processing huge queryset, and also database error due to that the database pulls whole table at once. Concurrent database modification wouldn't make some entries repeated or skipped in this process.

This scares me that this is a workaround for getting an anti-pattern to not break things, rather than a best practice that we want people to use. Note that we shouldn't be adding this to the library if we don't think this is a best practice.

Again, I think @ormsbee is in a better position to help determine if this should be a part of a best practice, or if not, how to suggest better alternatives.

robrap (commented):
Thank you.  As noted in my cmoments, I'd like to get @ormsbee involved.

edx_django_utils/db/queryset_utils.py
ormsbee: Where are we going to be using this to start?

robrap: @ormsbee: I'm not sure about new uses, but you can see existing uses here: https://github.com/search?q=org%3Aedx+chunked_queryset&type=code

ormsbee: @robrap: So it looks like this is being used in management commands, which makes sense.

I'm not super-thrilled by this code snippet in that it does its own `order_by`–it makes perfect sense given the implementation, but it's surprising behavior from the outside. I'd honestly just rather avoid using it if possible most of the time.

That being said, when memory usage actually is a problem (e.g. iterating through millions of records), this is about the only hammer to reach for. The normal solution for this is supposed to be [iterator()](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iterator), but that only really does the cursor thing with Postgres and Oracle. The MySQL driver loads it all into memory anyway.

So I'm fine with its inclusion in the library, provided that there is proper guidance on when to use it and the caveats (e.g. "it throws away your sorting"). I wouldn't want this casually used, but in the circumstances that it's needed, it's probably better implemented once than to have three or four reimplemented versions of this floating around different repos in the codebase.

ormsbee: I think we can be more aggressive with the default chunk size through. The default Django uses for its own `iterator()` method is 2000, so that's probably a decent starting point.

robrap: Thanks @ormsbee. 
@HammadAhmadWaqas - please update both the default and the docs accordingly based on the notes above. Thank you!

edx_django_utils/__init__.py
HammadAhmadWaqas: Thank you for sharing this document. Now updated version accordingly.

edx_django_utils/db/queryset_utils.py
HammadAhmadWaqas: I have updated the PR.

@robrap @ormsbee 
Thank you so much for both of you for this detailed and quick review. This will help a lot to ship the dependent feature on time.

HammadAhmadWaqas:
@robrap I have incorporated feedback. Could you please review it again?

edx_django_utils/db/__init__.py
```diff
@@ -2,8 +2,10 @@
 Utilities for working effectively with databases in django.
 
 read_replica: Tools for making queries from the read-replica.
+queryset_utils: Utils to used with Django QuerySets.
```
robrap, line 5:
```suggestion
queryset_utils: Utils to use with Django QuerySets.
```

edx_django_utils/db/queryset_utils.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+
+def chunked_queryset(queryset, chunk_size=2000):
+    """
+    Slice a queryset into chunks.
+
+    The function slices a queryset into smaller QuerySets containing chunk_size objects and then yield them.It is used
+    to avoid memory error when processing huge queryset, and also database error due to that the database pulls whole
+    table at once. Concurrent database modification wouldn't make some entries repeated or skipped in this process.
```
robrap, line 12: Here are some minor fixes.
```suggestion
    The function slices a queryset into smaller QuerySets containing chunk_size objects and then yields them. It is
    used to avoid memory error when processing huge querysets, and also to avoid database errors due to the
    database pulling the whole table at once. Concurrent database modification wouldn't make some entries
    repeated or skipped in this process.
```

However, I'm not clear on the meaning of the last sentence, so I'm not sure how to fix it.  Are you saying that without this function, another problem is that concurrent database modifications will cause some entries to be repeated or skipped? If so, maybe use:
```
Additionally, without using a chunked queryset, concurrent database modification while processing a large table might repeat or skip some entries.
```

Or, is this a problem that is introduced by chunked queryset?

robrap (approved):
Thanks.  Just some suggested cleanup for the comments.

Also, after the updates, can you copy/paste the final function docstring into the commit comment and/or PR description? Thanks!

edx_django_utils/db/queryset_utils.py
HammadAhmadWaqas: Yes, Without using this function, concurrent database modifications will cause some entries to be repeated or skipped, due to fetching all the data from the database at once.
I have added one more test case to support this case.

PR #109, opened 2021-03-24
Bump the version for `pluggable_decorator`

nedbat:
Update the version number to make pluggable_decorator available to users of this repo.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,15 @@ Change Log
 Unreleased
 ----------
 
+[3.16.0] - 2021-03-24
+---------------------
+
+Added
+_____
+
+* Added ``pluggable_override`` decorator.
+
+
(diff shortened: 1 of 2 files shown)


nedbat:
@Agrendalath let's get the version bumped.

Agrendalath:
@nedbat, good point, thank you for handling this!

Agrendalath:
@nedbat, could you please also create the `3.16.0` tag?

nedbat:
There's always another step... Done.

Agrendalath:
@nedbat, oh, I think it needs to be a release, not a tag. Cf. [the PyPi workflow](https://github.com/edx/edx-django-utils/blob/master/.github/workflows/pypi-publish.yml#L4).

nedbat:
Sorry! Done.

PR #118, opened 2021-04-30
ARCHBOM-1759: code owner squad and theme and other cleanup

robrap:
**Description:**

1. Add code_owner_squad and code_owner_theme custom attributes. See `0004-code-owner-theme-and-squad.rst` 
ADR for details on the decision.
2. The old location for `CodeOwnerMonitoringMiddleware` is being removed. It had moved in a past commit, and all
references in the Open edX platform have been updated to point to the new location. *BREAKING CHANGE*: But no one should be affected, since  no one is relying on this any more.
3. The happy-path behavior for the `CODE_OWNER_MAPPINGS` is untouched. This change now fails fast on
misconfigurations. *BREAKING CHANGE*: Since edX.org is relying heavily on the code_owner custom attributes, we will now fail fast, rather than just logging an error, when `CODE_OWNER_MAPPINGS` is misconfigured.

**Author concerns:**

This is our first reliance on the already configured `CODE_OWNER_THEMES` setting. The setting as it exists is a little strange, because instead of listing squad names under each theme, it lists 'theme-squad' names under each theme, so the theme needs to be parsed out to find the squad name.

There are two options:
1. Leave as-is and minimize implementation work. That is what was currently done.
2. Create a separate PR to update the output of the code that generates `CODE_OWNER_THEMES`, and wait for it to land and the jobs to automatically update this setting for all environments. Then, update this code to use the improved setting format.

[question] Reviewer: Should I take longer to make this cleaner? Once this is out, we probably won't fix it in the future.

**JIRA:**

[ARCHBOM-1759](https://openedx.atlassian.net/browse/ARCHBOM-1759)

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


edx_django_utils/monitoring/docs/how_tos/add_code_owner_custom_attribute_to_an_ida.rst
```diff
@@ -47,23 +53,29 @@ An untested potential alternative to the decorator is documented in the `Code Ow
...
         - openedx.core.djangoapps.xblock
-      team-blue:
+      theme-x-team-blue:
       - '*'  # IMPORTANT: you must surround * with quotes in yml
 
+    # YAML format of example CODE_OWNER_THEMES
+    CODE_OWNER_THEMES:
+      theme-x:
+      - theme-x-team-red
+      - theme-x-team-blue
```
robrap, line 74: **Note to Reviewer:** See "Author Concerns" in the PR description.

edx_django_utils/monitoring/tests/code_owner/test_middleware.py
```diff
@@ -53,28 +62,38 @@ def test_request_call(self):
...
     )
-    @patch('edx_django_utils.monitoring.internal.code_owner.middleware.set_custom_attribute')
+    @patch(
+        'edx_django_utils.monitoring.internal.code_owner.utils.set_custom_attribute',
+        new_callable=get_set_custom_attribute_mock
+    )
+    @patch(
+        'edx_django_utils.monitoring.internal.code_owner.middleware.set_custom_attribute',
+        new_callable=get_set_custom_attribute_mock
+    )
```
robrap, line 75: **Note for Reviewer**: This is ugly, but is it worth trying to create a decorator that wraps this? Not sure how easy or hard that will be.

dianakhuang (approved):
Discussed this in person and approved the current implementation.

edx_django_utils/monitoring/docs/how_tos/add_code_owner_custom_attribute_to_an_ida.rst
robrap: We're sticking with this to avoid additional work.

edx_django_utils/monitoring/tests/code_owner/test_middleware.py
robrap: We're sticking with this. It works.

PR #122, opened 2021-05-05
feat: Added upgrade-python-requirements workflow

Jawayria:
This workflow runs make upgrade at scheduled time and notifies if failure occurs

JIRA: https://openedx.atlassian.net/browse/BOM-2509


.github/workflows/upgrade-python-requirements.yml
```diff
@@ -0,0 +1,68 @@
+name: Upgrade Requirements
+
+on:
+  schedule:
+    # will start the job at 06:00 UTC every Monday
+    - cron: "0 6 * * 1"
```
mraarif, line 6: this one should be Okay I think since Lahore time would be UTC + 5 and (just an FYI) we want the upgrade PRs to be ready before workday for Lahore starts

mraarif (approved):

UsamaSadiq (approved):

PR #143, opened 2021-07-02
adding commandd.

awais786:
Some experiment.

### DIFF
--- a/edx_django_utils/admin/manage_users_by.py
+++ b/edx_django_utils/admin/manage_users_by.py
@@ -36,7 +36,7 @@ def is_valid_django_hash(encoded):
 class manage_users:  # lint-amnesty, pylint: disable=missing-class-docstring
 
     def __init__(self, name, email, password):
-        manage_users.Model.__init__(self)
+        super().__init__(*args, **kwargs)
         self.name = name
         self.email = email
         self.email = password
@@ -83,7 +83,7 @@ class manage_users:  # lint-amnesty, pylint: disable=missing-class-docstring
         user.delete()
 
     @transaction.atomic
-    def handle(self, username, email, is_remove, is_staff, is_superuser, groups,  # lint-amnesty, pylint: disable=arguments-differ
+    def update_user(self, username, email, is_remove, is_staff, is_superuser, groups,  # lint-amnesty, pylint: disable=arguments-differ
                unusable_password, initial_password_hash, *args, **options):
 
         if not UserProfile:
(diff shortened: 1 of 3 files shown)


robrap:
@awais786: _If this goes beyond an experiment_, can you add an explanation and also explain why this should live in this repo. No need to respond unless you push forward with this later. Thank you!

awais786:
> @awais786: _If this goes beyond an experiment_, can you add an explanation and also explain why this should live in this repo. No need to respond unless you push forward with this later. Thank you!

awais786:
@robrap I am checking how management command works from this package. But missing some thing. Package is installed in shell but not command not runs successfully.

regisb:
@awais786  I think you should move the `manage_users_by.py` file from the `admin/` to the `management/commands` folder.

awais786:
> @awais786 I think you should move the `manage_users_by.py` file from the `admin/` to the `management/commands` folder.

@regisb I am checking some thing. But If i continue with this I will move that there. Thanks.

edx_django_utils/admin/manage_users_by.py
```diff
@@ -0,0 +1,142 @@
...
+from django.contrib.auth.hashers import identify_hasher, is_password_usable
+from django.contrib.auth.models import Group
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.utils.translation import gettext as _
+
+logger = getLogger(__name__)  # pylint: disable=invalid-name
+
+
+try:
```
adzuci, line 18: Hey @awais786, thanks for putting this PR up!  I and Saad agree that the try catch statement is easier than writing a `post_save` signal handler, though it's not clear to me what the pros and cons are.

Do you think we can use the same technique for the generate password logic or somehow have an if statement that only hits that logic if in an edxapp app?

https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/management/commands/manage_user.py#L100-L110

e.g. 

```
try:
    from openedx.core.djangoapps.user_authn.utils import generate_password
except ImportError:
    UserProfile = None
```

awais786: @adzuci  Sorry I missed this msg. 
Yes you can do this as well. I don't see any issue. Its already implemented at several places. Please check this link.
https://github.com/edx/edx-enterprise/blob/3215aa03dfcee02cbb6d795fb29e025b4656398a/enterprise/utils.py#L45

PR #153, opened 2021-08-13
feat: adding function to query new relic using python

jinder1s:
**Description:**

Adding function to run a nrql query on new relic data. This was helpful in data gathering for cookie work. I suspect this will be useful in the future.

### DIFF
--- a/edx_django_utils/monitoring/scripts/run_nrql_on_nerd_graph.py
+++ b/edx_django_utils/monitoring/scripts/run_nrql_on_nerd_graph.py
@@ -1,0 +1,32 @@
+import requests
+import json
+import os
+
+
+def run_nrql_query(query, api_key, account_id):
+    """More details on this can be found at: https://developer.newrelic.com/collect-data/get-started-nerdgraph-api-explorer/"""
+    graphql_url = 'https://api.newrelic.com/graphql'
+    headers = {'Content-Type': 'application/json', 'API-Key': api_key}
+    graphql_query = f"""{{
+   actor {{
+      account(id: {account_id}) {{
+         nrql(query: "{query}") {{
+            results
+         }}
+      }}
+   }}
+}}"""
+    post_data = {'query': graphql_query.replace('\n','')}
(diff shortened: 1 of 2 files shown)


edx_django_utils/monitoring/scripts/run_nrql_on_nerd_graph.py
```diff
@@ -0,0 +1,36 @@
...
+   actor {{
+      account(id: {account_id}) {{
+         nrql(query: "{query}") {{
+            results
+         }}
+      }}
+   }}
+}}"""
+    post_data = {'query': graphql_query.replace('\n','')}
+    response = requests.post(url=graphql_url,  headers = headers, json=post_data)
```
robrap, line 24: Might results be paged like for other calls? If so, should we add a comment/warning that this script doesn't handle paging? See example here: https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/scripts/new_relic_nrql_search.py#L166-L167

robrap (commented):
I think this would be great to land, and I don't want to add a lot of work, but maybe a little TLC for documenting this, so people know how to use it and that it even exists.

1. [non-blocking] Consider using click for consistency and to provide usage information. See https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/scripts/new_relic_nrql_search.py.
2. Improve usage documentation in docstring. See https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/scripts/new_relic_nrql_search.py#L1-L11 for example.
3. Mention in docs. Although not perfect, maybe you could just add to https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/docs/how_tos/search_new_relic_nrql.rst? If you end up renaming this file to make it more general, please update references.
4. You can also add a note about this to the Unreleased section of the README without introducing a new version.

jinder1s:
Closing this cause I never got a chance to come back to this and I no longer remember the context for this PR.

PR #161, opened 2021-09-20
build: Updated the workflow to trigger release on tags

Jawayria:
JIRA: https://openedx.atlassian.net/browse/BOM-2840


mzulqarnain1 (approved):

UsamaSadiq (approved):

robrap:
Can someone explain this change to me. Does this better match how we did things with tox, or are we doing things differently? If differently, do we have an ADR or anything to explain the change? Thank you.

PR #188, opened 2022-01-24
fix: No longer clear RequestCache during exception phase (was too early)

timmc-edx:
Remove `RequestCacheMiddleware.process_exception` and add some notes about
what's actually necessary for proper cache cleanup.

Also clean up some changelog formatting issues and remove a heading for a
release that didn't happen.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


CHANGELOG.rst
```diff
@@ -3,19 +3,23 @@ Change Log
...
-   This project adheres to Semantic Versioning (http://semver.org/).
+   This project adheres to Semantic Versioning (https://semver.org/).
 
 .. There should always be an "Unreleased" section for changes pending release.
 
 Unreleased
 ----------
 
-[4.4.2] - 2021-12-23
---------------------
```
robrap, line 18: Should this be restored, and you could add `4.4.3` and update the version?

edx_django_utils/cache/middleware.py
```diff
@@ -18,16 +18,13 @@ def process_request(self, request):
...
-        return response
-
-    def process_exception(self, request, exception):    # pylint: disable=W0613
-        """
-        Clear the request cache after a failed request.
+        Clear the request cache after processing a response.
         """
+        # This is to just to do some basic cleanup, and isn't
+        # necessary for correctness; the next request will have the
+        # cache cleared during process_request anyhow.
```
robrap, line 25: [nit] Should we add something like: "Do NOT clear during process_exception, or it will cause issues for later process_response calls that still require the cache." This is non-blocking, so if you think the changelog is good enough as a warning, we can leave as-is.

CHANGELOG.rst
timmc-edx: There was never a 4.4.2 tag or release, so I could just restore it. Do we generally mark it in the changelog before it actually exists?

edx_django_utils/cache/middleware.py
timmc-edx: I was thinking that "not required for correctness" would be enough to keep someone who is overly cautious from restoring it, but I could add a note as well.

CHANGELOG.rst
robrap: I see. Generally, I would add to changelog and update version all in same PR if I plan to release. In this case, looks like `4.4.2`. 

I do not know why `4.4.2` had an entry here. Is the `Updated` section of `4.4.2` just a variation of `4.4.1`, or did additional work happen on the `urls` since `4.4.1`?

robrap (approved):
Just need to resolve the version update and changelog. Thanks!

CHANGELOG.rst
timmc-edx: It looks like someone made a change relating to django.urls and intended to release a 4.4.2 that day, but never got around to it. So this will end up being 4.4.2, in practice.

PR #209, opened 2022-05-04
feat: Add utility for logging sensitive information using encryption

timmc-edx:
**Description:**

This is copied from edx-platform, and was introduced there in
<https://github.com/openedx/edx-platform/pull/29682>. Bringing it out into
this library will allow it to be used across IDAs, and in particular in
the cookie monitoring middleware.

Introduces `log_sensitive` module for public-key encryption of sensitive
debug information in log messages, including CLI commands for generating
keys and decrypting log output. Exposes `log_sensitive` in logging module.

Changes from original:

- New CLI check to warn on inclusion of `[encrypted: ...]` wrapper,
  since I've made that mistake myself a few times. (Not sure if
  it's worth automating removal, since there are probably a few
  variations on it.)
- Update docstring to use correct module name (after move)
- Tweak to `gen-keys` language that I felt was a little confusing
- Test full CLI of log_sensitive (although had to mark as no-cover
  since sometimes coverage doesn't manage to detect what the
  subprocess code runs.)

Also:

- Expose as `log-sensitive` in CLI, since the code is now in an
  `internal` module.
- Add a Logging README (including for existing log filters code)
- Alphabetize main README's table of contents

ref: ARCHBOM-2082

**JIRA:**

[ARCHBOM-2082](https://openedx.atlassian.net/browse/ARCHBOM-2082)

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


timmc-edx:
Note for reviewers: Probably best reviewed commit-by-commit:

- First commit is a straight copy of the src and test files, plus a dependencies change
- Second commit makes some tweaks (see commit message)

robrap (commented):
Generally looks good. I'd like to see some minor refactoring and additional docs.
1. Looks like `logging` isn't mentioned in top-level README.rst.
2. Implementation should be moved to `logging/internal` and exposed through `__init__.py`.
3. The `logging` directory should have a README.rst with its purpose.

timmc-edx:
Ah right, I remember now that I had a question about that: If I just move this into internal and expose `encrypt_for_log`, then the CLI commands will have to call `edx_django_utils.logging.internal.log_sensitive` or something. But the CLI isn't internal... Thoughts on that?

timmc-edx:
Hmm, maybe I can make it a command/entry-point/whatever it's called, in setup.py.

Makefile
```diff
@@ -64,6 +64,7 @@ requirements: ## install development environment requirements
 	pip install -r requirements/pip.txt
 	pip install -qr requirements/pip-tools.txt
 	pip-sync requirements/dev.txt requirements/private.*
+	pip install . # CLI entry points
```
timmc-edx, line 67: This approach seems to be working OK. My first attempt of `pip install -e .` caused unpredictable failures -- the script stub that was created was different (and broken.) And coverage seems to have stopped being able to see inside the subprocess call, so I had to make some exclusions. But overall I guess the experience is nicer.

README.rst
```diff
@@ -28,6 +28,8 @@ This repository includes shared utilities for:
 
 .. _Monitoring Utilities: edx_django_utils/monitoring/README.rst
 
+.. _Logging Utilities: edx_django_utils/logging/README.rst
```
robrap, line 31: 1. This adds a link, but you don't reference the link in the text above.
2. Should we use alphabetical order? We mostly do in the text, except the last section was probably just added to the end. Not sure why you decided to add this link in this order.

robrap (commented):
Minor comment. Thanks for the updates.

README.rst
timmc-edx: Ah whoops, fixed! As for order, I just saw Monitoring and thought "hey, that's closest thematically, I'll put it after that". But now it's alphabetical, and right above Monitoring instead. :-)

timmc-edx: It's interesting that some doc linter didn't catch that. I bet it would have been caught in at least one of our projects.

robrap: I would have moved the final bullet and link which are still out of order, but I won’t block on that.

robrap (approved):
Nice job.

README.rst
timmc-edx: Ah, I can slip that in as well.

PR #213, opened 2022-05-12
build: Adding django40 testing in ci.

awais786:
Adding djang40 testing in ci


nedbat:
What is the status of this? Should it be moved forward, or closed?

PR #215, opened 2022-05-17
feat: Improvements to corrupt cookie logging

timmc-edx:
**Description:**

- Make independent of other cookie logging; no longer needs to meet cookie
  size threshold or sampling rate.
- New setting `UNUSUAL_COOKIE_SAMPLING_LOG_CHUNK` helps avoid truncated
  (non-decryptable) messages by splitting the output across multiple log
  messages.

**JIRA:**

ref: 2U ARCHBOM-2082

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

migrations, etc.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,17 @@ Change Log
 Unreleased
 ----------
 
+[4.9.0] - 2022-05-19
+--------------------
+
+Changed
+~~~~~~~
+
+* Corrupt cookie logging:
+
+  * Now independent of other cookie logging; no longer needs to meet cookie size threshold or sampling rate.
+  * New setting ``UNUSUAL_COOKIE_SAMPLING_LOG_CHUNK`` helps avoid truncated (non-decryptable) messages by splitting the output across multiple log messages.
+
 [4.8.1] - 2022-05-06
 --------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,7 +23,7 @@ Changed
 * Corrupt cookie logging:
 
   * Now independent of other cookie logging; no longer needs to meet cookie size threshold or sampling rate.
-  * New setting ``UNUSUAL_COOKIE_SAMPLING_LOG_CHUNK`` helps avoid truncated (non-decryptable) messages by splitting the output across multiple log messages.
+  * New setting ``UNUSUAL_COOKIE_HEADER_LOG_CHUNK`` helps avoid truncated (non-decryptable) messages by splitting the output across multiple log messages.
 
 [4.8.1] - 2022-05-06
 --------------------
--- a/edx_django_utils/monitoring/internal/middleware.py
+++ b/edx_django_utils/monitoring/internal/middleware.py
@@ -295,8 +295,10 @@ class CookieMonitoringMiddleware:
         """
         Add logging and custom attributes for monitoring cookie sizes.
 
-        Don't log contents of cookies because that might cause a security issue.
-        We just want to see if any cookies are growing out of control.
+        For cookie size monitoring, we don't log raw contents of cookies because that might
+        cause a security issue—we just want to see if any cookies are growing out of control.
+        However, there is also an option for encrypted logging of cookie data where there
+        appears to be some kind of data corruption occurring.
 
         Useful NRQL Queries:
 
@@ -317,7 +319,8 @@ class CookieMonitoringMiddleware:
 
             - COOKIE_HEADER_SIZE_LOGGING_THRESHOLD
             - COOKIE_SAMPLING_REQUEST_COUNT
-            - UNUSUAL_COOKIE_SAMPLING_PUBLIC_KEY
+            - UNUSUAL_COOKIE_HEADER_PUBLIC_KEY
+            - UNUSUAL_COOKIE_HEADER_LOG_CHUNK
 
         Returns: The message to be logged. This is returned, rather than directly
             logged, so that it can be processed at request time (before any cookies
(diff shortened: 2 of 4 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,17 @@ Change Log
...
+[4.9.0] - 2022-05-19
+--------------------
+
+Changed
+~~~~~~~
+
+* Corrupt cookie logging:
+
+  * Now independent of other cookie logging; no longer needs to meet cookie size threshold or sampling rate.
+  * New setting ``UNUSUAL_COOKIE_SAMPLING_LOG_CHUNK`` helps avoid truncated (non-decryptable) messages by splitting the output across multiple log messages.
```
robrap, line 26: For these two settings, why do they use the term `SAMPLING`? Is that just copy/paste, or because you think there might be other unusual cookies? I think I'd like `UNUSUAL_COOKIE_HEADER_XXX` better.

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -400,18 +408,71 @@ def get_log_message_and_monitor_cookies(self, request):
...
-
         return log_message
+
+    def log_corrupt_cookie_headers(self, request, corrupt_cookie_count):
+        """
+        Log all headers when corrupt cookies are detected (if settings permit).
+
+        This log data is encrypted using the log-sensitive utility.
+
+        - Logging requires that ``UNUSUAL_COOKIE_SAMPLING_PUBLIC_KEY`` is set.
```
robrap, line 419: See earlier comment about maybe using `UNUSUAL_COOKIE_HEADER_PUBLIC_KEY` instead.

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -400,18 +408,71 @@ def get_log_message_and_monitor_cookies(self, request):
...
+def split_ascii_log_message(msg, chunk_size):
+    """
+    Generator that splits a message string into chunks of at most ``chunk_size`` characters.
+
+    Message must consist of single-byte (ASCII-range) characters, otherwise chunks will
+    not reliably be capped at the ``chunk_size``.
+
+    A small collation tag will be added to the end of each chunk, and it may not be possible
+    to reliably predict the length of a log message's prefix (in its final output format), so
+    the ``chunk_size`` should be set conservatively.
```
robrap, line 462: Does conservatively mean 10, 100, 1000, etc.? Based on the default, for us it means 1000, so maybe you could provide 1000 characters as an example.

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -400,18 +408,71 @@ def get_log_message_and_monitor_cookies(self, request):
...
+
+        header_data = json.dumps(dict(request.headers.items()))
+        enc_output = encrypt_for_log(header_data, corrupt_cookie_log_pub_key)
+        msg = f"All headers for request with corrupted cookies (count={corrupt_cookie_count}): {enc_output}"
+
+        for piece in split_ascii_log_message(msg, chunk_size):
+            log.info(piece)
+
+
+def split_ascii_log_message(msg, chunk_size):
```
robrap, line 453: I may be missing something, but I was hoping if this were done, it could go into `edx_django_utils/logging/` and be used for any really long log message. Also, the setting could be something like `LONG_LOG_MESSAGE_CHUNK_SIZE`, and it would just need to be set once to match the environment.

Anyway, I don't want to block on this, but that is what I had in mind.

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -400,18 +408,71 @@ def get_log_message_and_monitor_cookies(self, request):
...
+        corrupt_cookie_log_pub_key = getattr(settings, 'UNUSUAL_COOKIE_SAMPLING_PUBLIC_KEY', None)
+
+        if not corrupt_cookie_log_pub_key:
+            return
+
+        # .. setting_name: UNUSUAL_COOKIE_SAMPLING_LOG_CHUNK
+        # .. setting_default: 9000
+        # .. setting_description: If necessary, logs data in chunks of this size, splitting across
+        #   multiple log messages. This should be set with your deployment's maximum log message
+        #   size in mind: Setting it too high may result in truncated messages, which will prevent
```
robrap, line 441:
```suggestion
        #   size in mind. Setting it too high may result in truncated messages, which will prevent
```

robrap (commented):
Looking forward to seeing those logs.

CHANGELOG.rst
timmc-edx: Oh, yeah, that's just left over from when it was gated on sampling! I'll rename. (And that means this will be off by default; I'll need to rename in the config repo as well.)

timmc-edx: I'm thinking `UNUSUAL_COOKIE_HEADER_{PUBLIC_KEY,LOG_CHUNK}`

edx_django_utils/monitoring/internal/middleware.py
timmc-edx: Conservatively relative to absolute max log line size. If the max is 10,000, then chunk size should be maybe 9,000.

(This is different from the previous max cookie size to log, since this is _after_ expansion due to Base64.)

edx_django_utils/monitoring/internal/middleware.py
timmc-edx: Yeah. I was hoping to do that as well, but there's some fiddly stuff here around splitting on characters vs bytes when it comes to multibyte characters.

- If you split on character count, then you may exceed the line size if there are multibyte characters
- If you split on byte count, then multibyte characters like "犬" run a risk of being split between the bytes

It ended up being too much of a rabbit hole, although I'd still like to clean it up to make it generally useful—or find some library that already does this. In the meantime, it should be sufficient for this known-character-range code.

CHANGELOG.rst
robrap: This line is still using the old name.

robrap (approved):
One more rename, but looks good. Other non-blocking comment to consider.

timmc-edx:
Thanks! Updated.

PR #218, opened 2022-05-25
WIP cookie consent API, starting with OneTrust

timmc-edx:


edx_django_utils/cookie_consent/README.rst
```diff
@@ -0,0 +1,25 @@
...
+Usage
+-----
+
+Call ``has_consented`` before setting cookies, passing it the request object and the category of cookie that would be set. Example::
+
+    from edx_django_utils.cookie_consent import has_consented
+
+    def set_referrer(request, ...):
+        if has_consented(request, 'tracking'):
+            request.set_cookie(...)
```
robrap, line 25: I wonder if a good API might be something like:
```
set_cookie_if_consented(request, 'tracking', ...):
```
1. It could return a true/false in case there is other related conditional code.
2. We could lint for `set_cookie` and ensure that  `set_cookie_if_consented` is always used instead.

Additionally, instead of a string like 'tracking', would it make sense to use an enum or different method names, like `set_tracking_cookie_if_consented`? Maybe the library would need to make the options pluggable, but if a string were used, it would be nice to validate it against some setting.

edx_django_utils/cookie_consent/README.rst
```diff
@@ -0,0 +1,25 @@
...
+
+"Cookie" is a shorthand here for various types of tracking and storage, but cookies are the main focus.
+
+Configuration
+-------------
+
+Set ``COOKIE_CONSENT_SOURCE`` to the module and name of a subclass of ``ConsentSource``, either one provided by edx-django-utils or from another package. This class will be responsible for answering whether the user has consented to certain categories of cookies. Known values:
+
+- ``edx_django_utils.cookie_consent.OnlyNecessary`` is the default and will only permit strictly necessary cookies.
+- ``edx_django_utils.cookie_consent.OneTrustSource`` will read and interpret the cookie set by OneTrust's consent UI.
```
robrap, line 14: Again, ok as an example, but it should be from a different library.

robrap (commented):
@timmc-edx: I'm not certain what you do or don't want reviewed. Let me know if you have other specific questions. Thanks.

edx_django_utils/cookie_consent/README.rst
timmc-edx: It would be nice if we could lint for it in a simple way, but I don't think we can.

- There are a bunch of ways to interact with cookies, both reading and writing, and with various attributes that can be set: http-only, secure, same-site, etc. We'd have to know a lot more about cookie APIs than we really need to, just to do mostly pass-through wrapping.
- I want to wrap other pieces of code, such as sending data to Segment or including script tags in templates.

edx_django_utils/cookie_consent/README.rst
timmc-edx: In general terms, edx-django-utils shouldn't have to know about specific consent cookie formats, even of the most popular solutions like OneTrust. But on the other hand, I'm not sure I can justify making a Python package that contains about 20 lines of code.

edx_django_utils/cookie_consent/internal/core.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+import re
+from functools import lru_cache
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+
+
+class ConsentSource:
```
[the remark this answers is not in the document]
timmc-edx: Agreed, it's a good place for those things once I get to that.

edx_django_utils/cookie_consent/README.rst
robrap: Ok. We could still do basic linting, even if it as simple as ensuring that `has_consented` has been imported into any module that is also importing `set_cookie`, etc. It wouldn't ensure full coverage of every usage within a model, but might catch some issues.

edx_django_utils/cookie_consent/README.rst
```diff
@@ -0,0 +1,25 @@
...
+Utilities for respecting user consent regarding different cookie categories.
+
+"Cookie" is a shorthand here for various types of tracking and storage, but cookies are the main focus.
+
+Configuration
+-------------
+
+Set ``COOKIE_CONSENT_SOURCE`` to the module and name of a subclass of ``ConsentSource``, either one provided by edx-django-utils or from another package. This class will be responsible for answering whether the user has consented to certain categories of cookies. Known values:
+
+- ``edx_django_utils.cookie_consent.OnlyNecessary`` is the default and will only permit strictly necessary cookies.
```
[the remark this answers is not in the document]
timmc-edx: Hmm... I suppose that when this lands in Olive, we probably want to start off with `AcceptAll`. But I think we probably want to _eventually_ use `OnlyNecessary` by default, since that simplifies compliance.

edx_django_utils/cookie_consent/README.rst
timmc-edx: Yeah, I think that might be as close as we can get.

Another interesting question is whether we wrap "strictly necessary" cookie code in this check even though it will always return True. I think the answer is yes, since it makes it easier to tell what cookies have been categorized. Thoughts?

edx_django_utils/cookie_consent/README.rst
robrap: I misread something and thought we were adding some sort of OneTrust dependency. I think this is probably fine. Maybe it could be in its own module (e.g. `cookie_consent.sources.onetrust`)? This is just a thought. Take it or leave it. :)

edx_django_utils/cookie_consent/README.rst
robrap: I'm not clear on what you are saying.

timmc-edx: There's a special category of "strictly necessary", which I've designated via the string `"necessary"` so far (although exact naming is still TBD). These are cookies and other storage activities that are required for the functioning of the site. An example would be the cookie `edx-jwt-cookie-header-payload`. Here's how it's set currently:

https://github.com/openedx/edx-platform/blob/432cde31a5b0c66acfb45d421c919cf362990259/openedx/core/djangoapps/user_authn/cookies.py#L331-L335

If someone is looking at that code, they might wonder "hmm, has this set-cookie been categorized yet?" So we could wrap it in `if has_consented(request, 'necessary'):` just to signal that it has indeed been categorized. If so, the `necessary` category needs to be understood as special (at the very least in the ConsentSource subclasses.)

Or, we can mark those locations with some kind of annotation, or even just not mark them at all, in which case the code in this module doesn't need to know about the `necessary` category, since nothing should ever ask.

robrap: Maybe it can be marked with something like?:
```
# necessary cookie, pylint: disable=missing-cookie-consent
```
I personally think the `has_consented` call that always returns True is a confusing indicator.

robrap: Or, as you mentioned, it would also work to have:
```
.. cookie_name: ...
.. cookie_category: necessary
.. cookie_description: ...
```

timmc-edx:
This work turned out not to be necessary; all of the scripts of concern are loaded by Javascript!

PR #239, opened 2022-09-08
docs: Relicense from AGPL 3 to Apache 2

timmc-edx:
See discussion on https://github.com/openedx/tcril-engineering/issues/289 for background (and authorization).
**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,674 +1,202 @@
 
-GNU AFFERO GENERAL PUBLIC LICENSE
-                       Version 3, 19 November 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
-
-  The GNU Affero General Public License is a free, copyleft license for
-software and other kinds of works, specifically designed to ensure
-cooperation with the community in the case of network server software.
-
-  The licenses for most software and other practical works are designed
-to take away your freedom to share and change the works.  By contrast,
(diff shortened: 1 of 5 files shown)


LICENSE.txt
```diff
@@ -1,674 +1,202 @@
...
-of the code.  There are many ways you could offer source, and different
-solutions will be better for different programs; see section 13 for the
-specific requirements.
-
-  You should also get your employer (if you work as a programmer) or school,
-if any, to sign a "copyright disclaimer" for the program, if necessary.
-For more information on this, and how to apply and follow the GNU AGPL, see
-<http://www.gnu.org/licenses/>.
-
-EdX Inc. wishes to state, in clarification of the above license terms, that
```
timmc-edx, line 664: Of possible interest, note the custom addendum we had here. (Nothing similar should be required for Apache 2, as far as I am aware.)

setup.py
```diff
@@ -121,15 +121,14 @@ def is_requirement(line):
             'log-sensitive = edx_django_utils.logging.internal.log_sensitive:cli',
         ],
     },
-    license="AGPL 3.0",
```
timmc-edx, line 124: Apparently this field is only intended for use when you don't have a license with a known classifier, which we do in fact supply farther down.

LICENSE.txt
kdmccormick: Thanks for pointing that out, I'd never noticed that addendum. I agree that it doesn't matter for this PR, but it's good to know about.

kdmccormick (approved):
Looks good, thanks a bunch @timmc-edx .

Issue #241, opened 2022-09-12
Move new IP address code from edx-platform to edx-django-utils


timmc-edx:
Move the IP code created in https://2u-internal.atlassian.net/browse/ARCHBOM-2056 into edx-django-utils. This way it can be used by other IDAs, and more easily forked to an external library (e.g. replacement for django-ipware?) if someone outside of Open edX wants to do so.

A/C:

- https://github.com/openedx/tcril-engineering/issues/289
- https://github.com/openedx/edx-django-utils/pull/242
- https://github.com/openedx/edx-platform/pull/31000
- [x] 2U internal: Notify SWG that this has been done

Moved here from <https://openedx.atlassian.net/browse/ARCHBOM-2057>.

PR #242, opened 2022-09-13
feat: Move safe IP address handling code from edx-platform

timmc-edx:
This will allow using the code in all IDAs.

Part of <https://github.com/openedx/edx-django-utils/issues/241>.
Code will be removed from edx-platform in <https://github.com/openedx/edx-platform/pull/31000>.

----

Commits best reviewed separately.

----

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/edx_django_utils/ip/README.rst
+++ b/edx_django_utils/ip/README.rst
@@ -1,0 +1,17 @@
+IP Address Utils
+################
+
+Utilities for safely and correctly determining the IP address(es) of a request.
+
+See ``edx_django_utils/ip/__init__.py`` for documentation and a list of everything included in the public API.
+
+Usage
+*****
+
+The short version:
+
+1. Configure ``CLOSEST_CLIENT_IP_FROM_HEADERS``
+2. Make sure ``init_client_ips`` is called as early as possible in your middleware stack
+3. Call ``get_safest_client_ip`` whenever you want to know the caller's IP address
+
+For details, see ``__init__.py`` module docstring.
--- a/edx_django_utils/ip/internal/ip.py
+++ b/edx_django_utils/ip/internal/ip.py
@@ -1,0 +1,276 @@
+import ipaddress
+import warnings
+
+from django.conf import settings
+
+
+def _get_meta_ip_strs(request, header_name):
+    """
+    Get a list of IPs from a header in the given request.
+
+    Return the list of IPs the request is carrying on this header, which is
+    expected to be comma-delimited if it contains more than one. Response
+    may be an empty list for missing or empty header. List items may not be
+    valid IPs.
+    """
+    if not header_name:
+        return []
+
+    field_name = 'HTTP_' + header_name.replace('-', '_').upper()
+    header_value = request.META.get(field_name, '').strip()
+
+    if header_value:
+        return [s.strip() for s in header_value.split(',')]
+    else:
+        return []
+
+
+def get_raw_ip_chain(request):
+    """
+    Retrieve the full IP chain from this request, as list of raw strings.
+
+    This is uninterpreted and unparsed, except for splitting on commas and
+    removing extraneous whitespace.
+    """
+    return _get_meta_ip_strs(request, 'X-Forwarded-For') + [request.META['REMOTE_ADDR']]
+
+
+def _get_usable_ip_chain(request):
+    """
+    Retrieve the full IP chain from this request, as parsed addresses.
+
+    The IP chain is the X-Forwarded-For header, followed by the REMOTE_ADDR.
+    This list is then narrowed to the largest suffix that can be parsed as
+    IP addresses.
+    """
+    parsed = []
+    for ip_str in reversed(get_raw_ip_chain(request)):
+        try:
+            parsed.append(ipaddress.ip_address(ip_str))
+        except ValueError:
+            break
+    return list(reversed(parsed))
+
+
+def _remove_tail(elements, f_discard):
+    """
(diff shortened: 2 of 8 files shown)


edx_django_utils/ip/README.rst
```diff
@@ -0,0 +1,17 @@
+Request Utils
```
robrap, line 1: IP Utils?

edx_django_utils/ip/internal/ip.py
```diff
@@ -357,15 +357,16 @@ def _get_client_ips_via_trusted_header(request):
         if closest_client_ip := _get_trusted_header_ip(request, header_name, index):
             # The equality check in this predicate is why we use parsed IP
             # addresses -- ::1 should compare as equal to 0:0:0:0:0:0:0:1.
-            external_chain = _remove_tail(full_chain, lambda ip: ip != closest_client_ip)  # pylint: disable=cell-var-from-loop
+            # pylint: disable=cell-var-from-loop
+            external_chain = _remove_tail(full_chain, lambda ip: ip != closest_client_ip)
```
robrap, line 201: Nit: Why this change? Did it also have line too long? You could disable that as well. You can leave this, but I like getting this stuff off to the side.

edx_django_utils/ip/internal/tests/test_ip.py
```diff
@@ -132,7 +132,8 @@ def test_get_client_ips_via_xff(self, xff, remote_addr, expected_strs):
         request_meta = {k: v for k, v in request_meta.items() if v is not None}
         self.request.META = request_meta
 
-        assert [str(ip) for ip in ip._get_client_ips_via_xff(self.request)] == expected_strs  # pylint: disable=protected-access
+        # pylint: disable=protected-access
+        assert [str(ip) for ip in ip._get_client_ips_via_xff(self.request)] == expected_strs
```
robrap, line 136: Nit: Same comment about pylint positioning.

robrap (commented):
1. Can you update the PR description so it stands on its own? Maybe just repeat the title, if you don't feel the need to say more. (This is personal preference.)
2. What is the plan for test coverage? Should we do it as part of this PR, since it will be library code that has a larger scope? If not, we could add pragmas with a comment pointing to the ticket to complete this, like a temporary amnesty.
3. I did not pay much attention to code and tests that seemed to just be copied over from the original.

edx_django_utils/ip/internal/ip.py
timmc-edx: Yeah, pycodestyle E501 line-too-long, and I was having trouble figuring out how to disable _just_ that warning (rather than using `noqa`). I can take another look to see if there's a way to do that,

timmc-edx: OK, apparently `noqa` has to appear _first_ in the comment, and doesn't accept things like `E501` -- it just entirely disables all errors of any sort on that line.

edx_django_utils/ip/README.rst
timmc-edx: Ah yes, thank you -- that was an artifact from an earlier version of this branch when I was considering having this be a broader "request utils" module.

edx_django_utils/ip/internal/ip.py
robrap: Are you saying the recommendation to use `# noqa: E501, pylint: disable=line-too-long` doesn't work?

robrap: I’m just curious, but feel free to leave this on two lines. I don’t mean to get us lost in this. Also why I wish there was a doc for this, if it could be solved once.

timmc-edx:
I don't like repeating the PR title in the body. I can link in the pending edx-platform PR, though. Do you think there's more that needs to be said beyond that?

timmc-edx:
I'm going to try to improve coverage and add pragmas where it doesn't seem worth it.

robrap:
> I don't like repeating the PR title in the body. I can link in the pending edx-platform PR, though. Do you think there's more that needs to be said beyond that?

Nope. We just have different preferences. Your PR. Go with yours. 🙂

edx_django_utils/ip/internal/ip.py
```diff
@@ -0,0 +1,285 @@
...
+    Return the list of IPs the request is carrying on this header, which is
+    expected to be comma-delimited if it contains more than one. Response
+    may be an empty list for missing or empty header. List items may not be
+    valid IPs.
+    """
+    if not header_name:
+        return []
+
+    field_name = 'HTTP_' + header_name.replace('-', '_').upper()
+    header_value = request.META.get(field_name, '').strip()
```
timmc-edx, line 28: I'm not sure why I'm not using `request.headers` here, and instead seem to have duplicated the logic that that uses. I wonder if it's possible that this was to work around the "request" objects we get sometimes being `django.http.request.HttpRequest` and sometimes being a DRF wrapper? But the DRF wrapper does proxy through any "missing" attributes via a dynamic `__getattr__`, so `headers` is available. Maybe I thought it wasn't.

Anyway, I'm going to see if I can just replace this code with `request.headers.get(field_name, '').strip()`.

timmc-edx:
Rewrote branch into two commits: One that's basically a straight copy, and one that makes new changes.

robrap (approved):

timmc-edx:
(rebased onto master)

timmc-edx:
(adjusted changelog date)

PR #246, opened 2022-10-06
feat: Add wrapper for New Relic's background_task.

dianakhuang:
We would like to be able to add our own custom handling to background task monitoring, so we added a simple wrapper around it.

### DIFF
--- a/edx_django_utils/monitoring/internal/utils.py
+++ b/edx_django_utils/monitoring/internal/utils.py
@@ -96,3 +96,21 @@ def record_exception():
     """
     if newrelic:  # pragma: no cover
         newrelic.agent.record_exception()
+
+
+def background_task(*args, **kwargs):
+    """
+    Handles monitoring for background tasks that are not passed in through the web server like
+    celery and event consuming tasks.
+
+    For more details, see:
+    https://docs.newrelic.com/docs/apm/agents/python-agent/supported-features/monitor-non-web-scripts-worker-processes-tasks-functions
+
+    """
+    def noop_decorator(func):
+        return func
+
+    if newrelic:  # pragma: no cover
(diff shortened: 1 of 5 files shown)


edx_django_utils/monitoring/internal/utils.py
```diff
@@ -96,3 +96,16 @@ def record_exception():
...
+    """
+    Handles monitoring for background tasks that are not passed in through the web server like
+    celery and event consuming tasks.
+
+    For more details, see:
+    https://docs.newrelic.com/docs/apm/agents/python-agent/supported-features/monitor-non-web-scripts-worker-processes-tasks-functions
+
+    """
+    if newrelic:  # pragma: no cover
+        newrelic.agent.background_task(*args, **kwargs)
```
robrap, line 111: Ned says add a return. Good call.

timmc-edx, line 111: Here's my best guess, untested...

```suggestion
    if newrelic:  # pragma: no cover
        return newrelic.agent.background_task(*args, **kwargs)
    else:
        return lambda f: f
```

dianakhuang: Yup, I was queueing up the change, and I think you're right.

timmc-edx: Hah! I'll be pleased if I was, but not surprised if I wasn't. :-P Despite being pretty comfortable with functional programming, decorators always somehow throw me for a loop. (Becca linked to it as a puzzle.)

dianakhuang: The parts where it gets translated from functional programming to syntactic sugar always trips me up, too.

rgraber (approved):
LGTM

PR #289, opened 2023-04-05
feat: Middleware to add Content-Security-Policy headers

timmc-edx:
This replaces https://github.com/edx/edx-arch-experiments/pull/256 and removes some features. First commit is a straight copy of the Python (plus a couple of unrelated fixes), second commit contains unreviewed changes.

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed


rgraber (approved):
LGTM

edx_django_utils/security/csp/middleware.py
```diff
@@ -0,0 +1,122 @@
...
+
+from django.conf import settings
+from django.core.exceptions import MiddlewareNotUsed
+
+
+def _load_headers() -> dict:
+    """
+    Return a dict of headers to append to every response, based on settings.
+    """
+    # .. setting_name: CSP_ENFORCE
```
timmc-edx, line 19: Author's note: I wonder if we want a less generic prefix than `CSP_` for these. Some of these might conflict with Mozilla's `django-csp` library.

timmc-edx: @rgraber What do you think about `CSP_STATIC_` as a prefix? (`CSP_STATIC_ENFORCE` etc.) This would help avoid collisions but would also highlight that these are static values, and cannot be used for the more advanced `strict-dynamic` CSP features.

timmc-edx:
Tests were passing before the previous two commits, but now there's something _badly_ wrong with the codecov package (https://pypi.org/project/codecov/) and CI is broken. But `make test`, `make quality`, and `make docs` all work locally, so I'm going to go ahead and merge.

Issue #288, opened 2023-04-05
Non-existent SettingsType referenced how-to documentation


rgraber:
The how-to for creating a new plugin app involves using the SettingsType enum, which does not exist. We should either add the enum or remove the references to it in the documentation.

rgraber:
Replace SettingsType.PRODUCTION and SettingsType.COMMON in the documentation with the actual strings they refer to and reference that they can be found in edx-platform here: https://github.com/openedx/edx-platform/blob/599d5d054cd43b3ec7a291e38de96108dfa28e14/openedx/core/djangoapps/plugins/constants.py#L25

PR #290, opened 2023-04-06
docs: remove references to SettingsType

rgraber:
The SettingsType enum is not present in edx-django-utils. Remove references to it and explain that the values are available in edx-platform. Also explain that other IDAs may require other values.

No version upgrade since this is just docs.

Issue: https://github.com/openedx/edx-django-utils/issues/288


edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst
```diff
@@ -32,12 +32,13 @@ file::
...
 
-4. configure the Plugin App in their AppConfig
+4. configure the Plugin App in their AppConfig. Note that in this example, we are explicitly configuring plugins for use in edx-platform. If your plugin is going to be used in another IDA, you may have different project and settings types. You will need to look at the IDA in question for what values it expects. You may want to add new values to the relevant enums.
+
 class::
 
    from django.apps import AppConfig
    from edx_django_utils.plugins.constants import (
-       ProjectType, SettingsType, PluginURLs, PluginSettings, PluginContexts
+       ProjectType, PluginURLs, PluginSettings, PluginContexts
```
robrap, line 41: Actually, if we are going to leave this as an LMS example, why not leave `ProjectType` and `SettingsType` as-is, except moving them to a separate import that is edx-platform specific. You might even move your note to be a code comment against the platform specific import statement. Thoughts?

rgraber: I'm not sure we want to encourage importing from edx-platform

robrap: Yes. I forgot this is plugin code. I was confused by `ProjectType` being left as-is. Should that also be converted to strings?

robrap: Also, I wonder if the ultimate solution is to just move all the constants to the library? What is the down side of that? Not sure if we'd need LMSSettingsType, etc. Anyway, I know that affects scope of this, but it is also difficult to use the example code if we leave a bunch of strings and it isn't clear where they are coming from.

rgraber: The downside that I can see is the precedent of having IDA-specific code in plugins that should theoretically be able to go in any IDA. I don't think we'd want to add constants every time we enabled plugins in a new IDA that had its own configuration vars.

rgraber: For now I'd just as soon remove ProjectType, although it actually appears in a few more places in the documentation, so I should remove it there too.

robrap: Sure. Maybe just add a code comment next to the first of the strings pointing to `ProjectType` so someone knows where they are coming from?

edx_django_utils/plugins/plugin_contexts.py
```diff
@@ -19,8 +19,10 @@ def get_plugins_view_context(project_type, view_name, existing_context=None):
     selected function to get their context dicts.
 
     Params:
-        project_type: a string that determines which project (lms or studio) the view is being called in. See the
-            ProjectType enum in plugins/constants.py for valid options
+        project_type: a string that determines which project the view is being called in. See the
+            ProjectType enum in
+            https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/plugins/constants.py#L14 for
```
robrap, line 24: 1. Also make this a permalink.
2. It is strange that there isn't an equivalent doc needed somewhere for the SettingsType.

edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst
robrap: Doesn't `ProjectType` also need to be dropped from the import?

edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst
```diff
@@ -71,17 +74,19 @@ class::
            # Configuration setting for Plugin Settings for this app.
            PluginSettings.CONFIG: {
 
-               # Configure the Plugin Settings for each Project Type, as needed.
-               ProjectType.LMS: {
+               # Configure the Plugin Settings for each Project Type, as needed. The full list setting types for edx-platform is
```
robrap, line 77:
```suggestion
               # Configure the Plugin Settings for each Project Type, as needed. The full list of setting types for edx-platform is 
```

edx_django_utils/plugins/plugin_contexts.py
```diff
@@ -19,8 +19,11 @@ def get_plugins_view_context(project_type, view_name, existing_context=None):
     selected function to get their context dicts.
 
     Params:
-        project_type: a string that determines which project (lms or studio) the view is being called in. See the
-            ProjectType enum in plugins/constants.py for valid options
+        project_type: a string that determines which project the view is being called in. See the
+            ProjectType enum in
+            https://github.com/openedx/edx-platform/blob/2dc79bcab42dafed2c122eb808cdd5604327c890/openedx/core/
+            djangoapps/plugins/constants.py#L14 .
+            for valid options for edx-platform
```
robrap, line 26:
```suggestion
            djangoapps/plugins/constants.py#L14
            for valid options for edx-platform.
```

edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst
```diff
@@ -71,17 +75,20 @@ class::
            # Configuration setting for Plugin Settings for this app.
            PluginSettings.CONFIG: {
 
-               # Configure the Plugin Settings for each Project Type, as needed.
-               ProjectType.LMS: {
+               # Configure the Plugin Settings for each Project Type, as needed. The full list of setting types for edx-platform is
+               # here:
+               # https://github.com/openedx/edx-platform/blob/2dc79bcab42dafed2c122eb808cdd5604327c890/openedx/core/djangoapps/plugins/constants.py#L25
```
robrap, line 80: Consider adding a period, as above.

robrap (approved):
Minor issues with periods, but looks good. Thank you!

PR #306, opened 2023-05-02
fix: enable updating pip.txt in make upgrade

UsamaSadiq:
## Description
- `pip.txt` wasn't being updated in the `make upgrade` job because the `--upgrade` flag was missing in the `pip-compile` command
- It was causing the scheduled `Python Requirements Upgrade` [build](https://github.com/openedx/edx-django-utils/actions/runs/4847705622/jobs/8666288926#step:5:28) to fail due to `pip & pip-tools` conflict.


zubairshakoorarbisoft (approved):
LGTM 👍

PR #330, opened 2023-07-28
fix: treat none value as cache miss

mumarkhan999:
**Description**
- We are replacing [(issue link)](https://github.com/edx/edx-arch-experiments/issues/92) `Python-memcache` with `pymemcache`.
- While doing that [(PR link)](https://github.com/edx/edx-internal/pull/8896) we faced the following error
```
  File "/edx/app/edxapp/edx-platform/xmodule/video_block/video_block.py", line 244, in student_view
    fragment = Fragment(self.get_html())
  File "/edx/app/edxapp/edx-platform/xmodule/video_block/video_block.py", line 355, in get_html
    branding_info = BrandingInfoConfig.get_config().get(user_location)
  File "/edx/app/edxapp/edx-platform/lms/djangoapps/branding/models.py", line 54, in get_config
    return json.loads(info.configuration) if info.enabled else {}
AttributeError: 'NoneType' object has no attribute 'enabled'
```
- The reason for the error is that at the following line [(line link)](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/branding/models.py#L54), the `info` variable is `None`
- This value for `info` is coming from `Tiered cache` [(link)](https://github.com/openedx/django-config-models/blob/master/config_models/models.py#L133)
- The Tiered cache is returning `None` (using `pymemcache` as backend) [(link)](https://github.com/openedx/edx-django-utils/blob/v5.5.0/edx_django_utils/cache/utils.py#L261). This `None` causes the exception.
- Previously, it wasn't creating any issue as `python-memcache` was treating `None` as `CACHE_MISS`
- Now, in this PR, we are explicitly treating `None` as `cache miss`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,8 +15,14 @@ Unreleased
 ----------
 
 Added
-
+~~~~~
 * Support added for Django 4.2
+
+Fixed
+~~~~~
+* Fixed bug where None was not properly being stored by TieredCache.
+  For backward compatibility, ``edx_django_utils.cache.disable_forced_cache_miss_for_none`` waffle switch has
+  been added, which defaults to the old broken behavior of treating None as a cache miss.
 
 [5.6.0] - 2023-07-24
 --------------------
--- a/edx_django_utils/cache/tests/test_utils.py
+++ b/edx_django_utils/cache/tests/test_utils.py
@@ -177,8 +177,9 @@ class TestTieredCache(DjangoTestCase):  # pylint: disable=missing-class-docstrin
         cached_response = self.request_cache.get_cached_response(TEST_KEY)
         self.assertTrue(cached_response.is_found, 'Django cache hit should cache value in request cache.')
 
+    @mock.patch('edx_django_utils.monitoring.internal.utils.set_custom_attribute')
     @override_switch('edx_django_utils.cache.disable_forced_cache_miss_for_none', True)
-    def test_get_cached_response_hit_with_cached_none(self):
+    def test_get_cached_response_hit_with_cached_none(self, mock_set_custom_attribute):
         """
         Tests cache hit when caching a None.
 
@@ -197,7 +198,13 @@ class TestTieredCache(DjangoTestCase):  # pylint: disable=missing-class-docstrin
         self.assertTrue(cached_response.is_found)
         self.assertEqual(cached_response.value, None)
 
-    def test_get_cached_response_miss_with_cached_none(self):
+        expected_calls = [
+            mock.call('retrieved_cached_none', True)
+        ]
+        mock_set_custom_attribute.assert_has_calls(expected_calls)
+
+    @mock.patch('edx_django_utils.monitoring.internal.utils.set_custom_attribute')
+    def test_get_cached_response_miss_with_cached_none(self, mock_set_custom_attribute):
         TieredCache.set_all_tiers(TEST_KEY, None)
         # Test retrieval from tier 1: RequestCache
         cached_response = TieredCache.get_cached_response(TEST_KEY)
@@ -208,6 +215,11 @@ class TestTieredCache(DjangoTestCase):  # pylint: disable=missing-class-docstrin
         # Test retrieval from tier 2: Django Cache
         cached_response = TieredCache.get_cached_response(TEST_KEY)
         self.assertFalse(cached_response.is_found)
+
+        expected_calls = [
+            mock.call('retrieved_cached_none', True)
+        ]
+        mock_set_custom_attribute.assert_has_calls(expected_calls)
 
     @mock.patch('django.core.cache.cache.get')
     def test_get_cached_response_force_cache_miss(self, mock_cache_get):
--- a/edx_django_utils/cache/tests/test_utils.py
+++ b/edx_django_utils/cache/tests/test_utils.py
@@ -6,6 +6,8 @@ from threading import Thread
 from unittest import TestCase, mock
 
 import ddt
+from django.test import TestCase as DjangoTestCase
+from waffle.testutils import override_switch
 
 from edx_django_utils.cache.utils import (
     DEFAULT_REQUEST_CACHE_NAMESPACE,
@@ -142,7 +144,7 @@ class TestRequestCache(TestCase):  # pylint: disable=missing-class-docstring
             RequestCache(DEFAULT_REQUEST_CACHE_NAMESPACE)
 
 
-class TestTieredCache(TestCase):  # pylint: disable=missing-class-docstring
+class TestTieredCache(DjangoTestCase):  # pylint: disable=missing-class-docstring
     def setUp(self):
         super().setUp()
         self.request_cache = RequestCache()
@@ -160,6 +162,13 @@ class TestTieredCache(TestCase):  # pylint: disable=missing-class-docstring
 
     @mock.patch('django.core.cache.cache.get')
     def test_get_cached_response_django_cache_hit(self, mock_cache_get):
+        """
+        Temporary tests for cache miss when caching a None.
+
+        Temporarily we are forcing cache misses by default when caching None for
+        backward compatibility purposes. See ``disable_forced_cache_miss_for_none``
+        switch for more details.
+        """
         mock_cache_get.return_value = EXPECTED_VALUE
         cached_response = TieredCache.get_cached_response(TEST_KEY)
         self.assertTrue(cached_response.is_found)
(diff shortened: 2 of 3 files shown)


awais786 (approved):

awais786 (approved):

robrap (changes requested):
Won’t this be a breaking change for cases where None was being cached?

robrap:
[idea] We could add some observability for when and if None is actually stored as a value. This won’t find all cases, but might help a transition.

mumarkhan999:
> Won’t this be a breaking change for cases where None was being cached?

Currently, the Memcache backend is `python-memcache`. If we cache `None`, it treats it as `CACHE_MISS`. So the functionality we're adding in this `PR` is never gonna executed as long as the backend is `python-memcache`. 

But using `pymemcache` as backend, the None will not be treated as `CACHE_MISS`. Instead, it will be returned as a valid value. And when we will try to retrieve items from `None` it will raise errors.

So what we're trying to do in this PR is keeping up the old functionality (treating None as CACHE_MISS) so that after updating the backend from `python-memcache` to `pymemcache` things keep running smoothly.

mumarkhan999:
> [idea] We could add some observability for when and if None is actually stored as a value. This won’t find all cases, but might help a transition.

Yeah, adding some observability might be a good thing. But the problem is not actually the `None` being cached. The problem is the way both backends (python-memcache & pymemcache) treat `None` upon `retrieval`. As long as we keep the behavior similar for both of them, there won't be any issues.

If there occur some errors because `None` is being treated as a valid value we'll find it through the `real errors`. If `None` isn't causing any `error` I guess we can ignore it. In our case, the actual errors are more important. Adding observability might be not much helpful here.

But if you still think it should be added. We can add it. Your call.

robrap:
@mumarkhan999: Thanks. The case I was confused about is this test for setting the cache with a None: https://github.com/openedx/edx-django-utils/blob/f9836a1fdc051f502ede86d13b465ce1f6d87hat90d/edx_django_utils/cache/tests/test_utils.py#L51-L55.  

It seems like your change isn't breaking that test, but I'm not sure how this would happen in the first place, given the test case I just linked to.

Apologies if I am missing something obvious here, but maybe this highlights my confusion?

Also, we may not need monitoring as you mentioned. First, let's see if I can better understand how this would ever happen.

Thank you.

mumarkhan999:
> @mumarkhan999: Thanks. The case I was confused about is this test for setting the cache with a None: https://github.com/openedx/edx-django-utils/blob/f9836a1fdc051f502ede86d13b465ce1f6d87hat90d/edx_django_utils/cache/tests/test_utils.py#L51-L55.
> 
> It seems like your change isn't breaking that test, but I'm not sure how this would happen in the first place, given the test case I just linked to.
> 



Actually, we only updated the `django-cache` layer of the `Tiered cache`. We are not doing anything with the `Request Cache` layer. Request Cache is `in-mem` python dict. We don't need to update anything in that layer. Hope, it answers your question.

robrap:
Thoughts:
1. I didn't realize that the test [test_get_cached_response_hit_with_cached_none](https://github.com/openedx/edx-django-utils/blob/f9836a1fdc051f502ede86d13b465ce1f6d8790d/edx_django_utils/cache/tests/test_utils.py#L51-L55) was from `TestRequestCache`.
2. The way that `TieredCache` works, it has a two tiers: 1) a request cache, and 2) a django cache. There is an assumption that they would provide the same result. It is unfortunate that there was no test for `test_get_cached_response_hit_with_cached_none` in `TestTieredCache`, and it may be that you uncovered a bug in the existing TieredCache. It may be that instead of storing a `_CACHE_MISS` object, we would need to create and store a `_CACHED_NONE` object, to be converted to a `None`, if we wanted this to work the way it was intended to work.

Does this make sense?

robrap (commented):
I found trying to explain the temporary situation so much more complex than just fixing the bug, so here is a PR that you can merge in here when you are ready: https://github.com/openedx/edx-django-utils/pull/332. Feel free to update as you wish.

edx_django_utils/cache/utils.py
```diff
@@ -4,6 +4,7 @@
 import hashlib
 import threading
 
+import waffle  # pylint: disable=invalid-django-waffle-import
```
mumarkhan999, line 7: I used waffle instead of the Django setting as we're already using the waffle switch in this repo.

https://github.com/openedx/edx-django-utils/blob/f9836a1fdc051f502ede86d13b465ce1f6d8790d/edx_django_utils/monitoring/internal/middleware.py#L19

edx_django_utils/cache/tests/test_utils.py
```diff
@@ -168,6 +168,35 @@ def test_get_cached_response_django_cache_hit(self, mock_cache_get):
         cached_response = self.request_cache.get_cached_response(TEST_KEY)
         self.assertTrue(cached_response.is_found, 'Django cache hit should cache value in request cache.')
 
+    @mock.patch('edx_django_utils.cache.utils.waffle.switch_is_active')
```
robrap, line 171: See https://waffle.readthedocs.io/en/stable/testing/automated.html for `override_switch` decorator for testing.

edx_django_utils/cache/utils.py
```diff
@@ -259,6 +260,10 @@ def _get_cached_response_from_django_cache(key):
             return CachedResponse(is_found=False, key=key, value=None)
 
         cached_value = django_cache.get(key, _CACHE_MISS)
+
+        if cached_value is None and not TieredCache._is_forced_cache_miss_for_none_disabled():
```
robrap, line 264: Where did my `_CACHED_NONE` code go? It seemed like simple code to make handling of caching `None` consistent for any memcached library, so I'd like to see it restored. Additionally, when the cached value is `_CACHED_NONE` on retrieval, I wanted to see something like `set_custom_attribute('retrieved_cached_none, True)`, regardless of the value of the switch.

edx_django_utils/cache/tests/test_utils.py
```diff
@@ -168,6 +168,35 @@ def test_get_cached_response_django_cache_hit(self, mock_cache_get):
...
+        self.assertEqual(cached_response.value, None)
+
+        self.request_cache.clear()
+        # Test retrieval from tier 2: Django Cache
+        cached_response = TieredCache.get_cached_response(TEST_KEY)
+        self.assertTrue(cached_response.is_found)
+        self.assertEqual(cached_response.value, None)
+
+    @mock.patch('edx_django_utils.cache.utils.waffle.switch_is_active')
+    def test_get_cached_response_miss_with_cached_none(self, mock_switch_is_active):
```
robrap, line 187: Maybe a docstring like the following:
```
Temporary tests for cache miss when caching a None.

Temporarily we are forcing cache misses by default when caching None for 
backward compatibility purposes. See ``disable_forced_cache_miss_for_none``
for more details.
```

CHANGELOG.rst
```diff
@@ -15,8 +15,10 @@ Unreleased
 ----------
 
 Added
-
+~~~~~
 * Support added for Django 4.2
+* ``edx_django_utils.cache.disable_forced_cache_miss_for_none`` waffle switch has been added to make
```
robrap, line 20:
```suggestion

Fixed
~~~~

* Fixed bug where None was not properly being stored by TieredCache. For backward compatibility, ``edx_django_utils.cache.disable_forced_cache_miss_for_none`` waffle switch has been added, which defaults to the old broken behavior of treating None as a cache miss.
```

robrap (commented):
Thanks.

edx_django_utils/cache/tests/test_utils.py
mumarkhan999: Thanks for mentioning it.
The reason I didn't use it at first because our `TestTieredCache` is inherited from `unittest.Testcae` which doesn't allow the database access and we get the following error when we try to check the `toggle` status.

```
*** RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.
```

BUT I've changed the parent class. It's working fine now.

robrap:
@mumarkhan999: Also, I should have mentioned this earlier, but during reviews it is nice to push new commits, rather than constantly squashing. If you use a commit comment like `squash! updated ...`, then commit lint won't complain, and you'll know what to squash later. This is nicer for reviews, because I can see just what has changed. Thanks.

edx_django_utils/cache/utils.py
mumarkhan999: `_CACHED_NONE` functionality has been added again. 

`set_custom_attribute('retrieved_cached_none, True)` was removed because it was giving `cyclic-import` related errors. I would add it in a `separate` commit.

edx_django_utils/cache/tests/test_utils.py
```diff
@@ -168,6 +168,35 @@ def test_get_cached_response_django_cache_hit(self, mock_cache_get):
         cached_response = self.request_cache.get_cached_response(TEST_KEY)
         self.assertTrue(cached_response.is_found, 'Django cache hit should cache value in request cache.')
 
+    @mock.patch('edx_django_utils.cache.utils.waffle.switch_is_active')
+    def test_get_cached_response_hit_with_cached_none(self, mock_switch_is_active):
```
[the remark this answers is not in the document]
mumarkhan999: Added.

edx_django_utils/cache/tests/test_utils.py
mumarkhan999: Added.

CHANGELOG.rst
mumarkhan999: Added.

edx_django_utils/cache/utils.py
```diff
@@ -303,6 +308,12 @@ def _should_force_django_cache_miss(cls):
         cached_response = DEFAULT_REQUEST_CACHE.get_cached_response(SHOULD_FORCE_CACHE_MISS_KEY)
         return False if not cached_response.is_found else cached_response.value
 
+    @staticmethod
+    def _is_forced_cache_miss_for_none_disabled():
+        # This switch is to provide backward compatibility.
+        # As we have been treating None as CACHE_MISS.
```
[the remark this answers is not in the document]
mumarkhan999: > Toggles are documented as described in these docs: https://edx.readthedocs.io/projects/edx-toggles/en/latest/how_to/documenting_new_feature_toggles.html#boilerplate-template.

Added the description. I just copied this previously

https://github.com/openedx/edx-django-utils/pull/70/files#diff-a54b48ac8b965a8dfa9c454409651a6312701e0c2b0df1341f835fab08ac65e3R222

edx_django_utils/cache/utils.py
```diff
@@ -259,6 +264,12 @@ def _get_cached_response_from_django_cache(key):
             return CachedResponse(is_found=False, key=key, value=None)
 
         cached_value = django_cache.get(key, _CACHE_MISS)
+
+        if cached_value == _CACHED_NONE and not TieredCache._is_forced_cache_miss_for_none_disabled():
+            cached_value = _CACHE_MISS
+        elif cached_value == _CACHED_NONE:
+            cached_value = None
```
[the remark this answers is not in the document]
robrap: You may need to import directly from internal. See untested import below:
```
from edx_django_utils/monitoring/internal/utils import set_custom_attribute
```

robrap (approved):
Looks great. Thanks.

PR #345, opened 2023-09-18
docs: Update Celery code owner ADR with failed attempt

timmc-edx:


edx_django_utils/monitoring/docs/decisions/0003-code-owner-for-celery-tasks.rst
```diff
@@ -29,8 +29,8 @@ Consequences
 (Rejected) Alternatives
 -----------------------
 
-An untested potential alternative to the ``@set_code_owner_attribute`` decorator is to try celery's `task_prerun signal`_ in an IDA, which would also ensure all future celery tasks are automatically handled. Although this is a potentially superior solution, it was missed at the time of implementation. We will not be switching to this solution at this time given we have a working solution and other priorities, but it is a potentially viable solution if the need arises.
+One hoped-for alternative we discovered after the initial ADR was to try setting the code owner in celery's `task_prerun signal`_ in an IDA, which would also ensure all future celery tasks are automatically handled. We `trialed the task_prerun approach <https://github.com/openedx/edx-platform/pull/33180>`_ but discovered that no attribute was set on the Celery tasks. This seems to be because New Relic instruments the task function itself with transaction-start and transaction-end calls, so any signals that are run before or after the task execution occur outside the scope of the transaction.
```
rgraber, line 32: I think it would be clearer to say something like
```suggestion
Setting the code owner in celery's `task_prerun signal`_ in an IDA

This would theoretically also ensure all future celery tasks in the IDA are automatically handled. We `trialed the task_prerun approach <https://github.com/openedx/edx-platform/pull/33180>`_ but discovered that no attribute was set on the Celery tasks. This seems to be because New Relic instruments the task function itself with transaction-start and transaction-end calls, so any signals that are run before or after the task execution occur outside the scope of the transaction.
```
And then add a change log or something to mention that this alternative was updated after the ADR was written.

rgraber (commented):
One suggestion for clarity.
Also GitHub isn't letting me comment on this line, but we should update the line "* New celery tasks will need to add the same decorator. See "Rejected Alternatives" for a potential alternative." to reflect that we no longer have a good alternative

edx_django_utils/monitoring/docs/decisions/0003-code-owner-for-celery-tasks.rst
timmc-edx: So, add a subheading?

rgraber: Sort of. Just state what the alternative is first before going into the history and the reasons for rejecting it. Doesn't have to be as a header.

rgraber (approved):
LGTM

PR #369, opened 2023-12-11
feat: Add manufacture_data django command

marlonkeating:
**Description:**

This change migrates the manufacture_data django command introduced in https://github.com/openedx/edx-enterprise/pull/1826 to where it can be used outside of the enterprise ecosystem.

**JIRA**

[ENT-7636](https://2u-internal.atlassian.net/browse/ENT-7636)

**Installation/Testing instructions:**

***In Devstack, prior to being pushed to PyPi***
1. Copy to IDA VM's python path
Example: `docker cp ~/edx-repos/edx-django-utils/. {container uuid}:/lib/`
Where `container uuid` can be found by running `docker ps`
2. Create management command that inherits from manufacture_data.command and imports model factories (Example: https://github.com/openedx/enterprise-catalog/pull/734)
3. Run manufacture_data management command
Example:
`python ./manage.py manufacture_data --model enterprise_catalog.apps.catalog.models.EnterpriseCatalog --title "Test Catalog"
`

**Merge checklist:**
- [ ] All reviewers approved
- [X] CI build is green
- [X] Version bumped
- [X] Changelog record added
- [X] Documentation updated (not only docstrings)
- [X] Commits are squashed

**Post merge:**
- [ ] Create a tag
- [ ] Check new version is pushed to PyPi after tag-triggered build is
      finished.
- [ ] Delete working branch (if not needed anymore)

PR #376, opened 2024-01-17
build: bump version

marlonkeating:
**Description:**

Follow-up from https://github.com/openedx/edx-django-utils/pull/369, where the version didn't get bumped in init.py


rgraber (approved):
LGTM

PR #384, opened 2024-02-13
build: adding 311 support.

awais786:
Python 312 support.

https://github.com/openedx/public-engineering/issues/233

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version: ['3.8', '3.12']
+        python-version: ['3.8', '3.11', '3.12']
         toxenv: [docs, quality, django42]
 
     steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -15,8 +15,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-20.04]
-        python-version: ['3.8']
-        toxenv: [docs, quality, django32, django42]
+        python-version: ['3.8', '3.11', '3.12']
```
robrap, line 18: @awais786: In order to save build time and resources, is it ok to only add 3.12 to start, and we can add 3.11 just to those that start failing? That's what some of us had in mind anyway. Let me know your thoughts.

robrap: That said, I guess it makes sense to experiment on one repo, because now we see that 3.11 is passing and that we just have a problem linting in 3.12. Carry on. We'll see what you learn. Thanks.

awais786: Got it, I will fix all the prs with python312 only.

irtazaakram (approved):

awais786:
https://github.com/openedx/axim-engineering/issues/1045

PR #402, opened 2024-03-29
build: Remove tests/ stub; make `make test-all` work right

timmc-edx:
- The `tests/__init__.py` create/delete thing seems to have been a workaround for something that's no longer relevant, so just remove it.
- The `test-all` target needed to depend on `clean` so that `build/lib` wouldn't stick around and confuse pytest, and the recipe needed to be pruned down to just `tox` because otherwise we'd run the quality checks twice. (`quality` is in the tox.ini default env list.)

**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,9 +69,6 @@ docs/edx_django_utils.*.rst
 requirements/private.in
 requirements/private.txt
 
-# tox environment temporary artifacts
-tests/__init__.py
-
 # Development task artifacts
 default.db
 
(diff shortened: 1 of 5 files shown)


robrap (approved):
Is https://github.com/openedx/edx-django-utils/blob/master/test_utils/__init__.py still relevant as well?

timmc-edx:
Good find -- I can remove test_utils as well.

PR #409, opened 2024-04-29
Feat  add dd and otel celery tasks

connorhaugh:
**Description:**

Adds an initialize_celery_monitoring util to allow celery tasks to be monitored by the Open Telemetry and Datadog providers. edx-platform PR to follow.

**JIRA:**

https://github.com/edx/edx-arch-experiments/issues/584

**Testing instructions:**

1. Set this package to be installed locally, check out this PR
2. 3. Run your edx-platform off this branch: https://github.com/openedx/edx-platform/pull/34673 to use the implementation required from the package.
3. set your OPENEDX_TELEMETRY django setting to be [`edx_django_utils.monitoring.DataDogBackend`]
4. follow https://2u-internal.atlassian.net/wiki/spaces/~840928901/pages/813793298/OpenTelemetry+New+Relic+and+Datadog for devstack setup for DD, but set up the DD agent on CMS, not lms
5. Follow https://2u-internal.atlassian.net/wiki/spaces/ENGAGE/pages/395673715/How+to+configure+Celery+worker+on+LMS+CMS+Devstack to set up celery workers on cms.
4. Export a course
6. View the export on DD with a celery task

**Reviewers:**
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

migrations, etc.


schenedx (approved):
👍

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -129,6 +156,9 @@ def record_exception(self):
         if span := self.dd_tracer.current_span():
             span.set_traceback()
 
+    def initialize_celery_monitoring(self, *args, **kwargs):
+        self.patch(celery=True)
```
timmc-edx, line 160: It looks like ddtrace patches Celery by default: https://github.com/DataDog/dd-trace-py/blob/01fbf9127180794392cc7dbe16acd610087dacf6/ddtrace/_monkey.py#L33 -- were you seeing otherwise?

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -103,6 +118,17 @@ def set_attribute(self, key, value):
...
 
+    def initialize_celery_monitoring(self, *args, **kwargs):
+        worker_process_init = kwargs.get('worker_process_init', None)
+        if worker_process_init is not None:
+            @worker_process_init.connect(weak=False)
+            def init_celery_tracing(*args, **kwargs):
+                self.instrumentor().instrument()
+        else:
+            raise Exception(
+                "the worker_process_init celery signal must be provided for OpenTelemetry to monitor celery tasks."
```
timmc-edx, line 129: If OpenTelemetry needs it, then it will have to be passed to all the backends, because there will only be one call site for the `initialize_celery_monitoring` API call we expose. However, it looks like it doesn't need to be passed in at all—this is just a signal, so we could `from celery.signals import worker_process_init`.

For that matter, it might make the most sense to instead say that the _caller_ should do that, like so:

```
@celery.signals.worker_process_init.connect(weak=False)
def init_celery_tracing(*args, **kwargs):
    edx_django_utils.monitoring.initialize_celery_monitoring()
```

We could also just leave the OpenTelemetry side of this unimplemented for now, since we won't be able to give it the testing it deserves (in production).

edx_django_utils/monitoring/internal/backends.py
robrap: I was just going by these docs: https://ddtrace.readthedocs.io/en/stable/integrations.html#celery.
This was in the ticket I had created: https://github.com/edx/edx-arch-experiments/issues/584
Maybe it was inaccurate?

Separately: In those docs I see that trace headers are off by default. We should find out if that is what we want from DD? Maybe long running tasks would kill our traces? Not sure. @connorhaugh: Can you task/track this?

connorhaugh:
Based on Tim's comment, I went and looked into if we were getting what we needed from celery by default.

To do that, I looked at the logs over the past month.

Do the logs related to celery increase or decrease in volume when we enabled DD in edx-platform?
[The answer is it increased in volume from april 22-27rd,](https://app.datadoghq.com/logs?query=service%3Acms-workers%20celery%20&agg_m=count&agg_m_source=base&agg_q=service&agg_q_source=base&agg_t=count&cols=&refresh_mode=sliding&storage=hot&top_n=10&top_o=top&viz=stream&x_missing=true&from_ts=1712667320393&to_ts=1715259320393&live=true) which was the window for edx-platform having celery on in dd. I have been really struggling to explain why that is the case, however, given the code that enables the patch by default. Either way, I'm making improvements to the code and we should maybe plan on needing it.

robrap:
@timmc-edx: FYI: I'm closing this PR. I assume at this point we do not need this.

PR #422, opened 2024-06-12
feat: add custom middleware for x-frame-options to allow overrides

jesperhodge:
**Description:**

The SCORM xblocks starting with openedx-scorm-xblock version 18 cannot load because they require the `X-Frame-Options` header to be set to `SAMEORIGIN` for scorm related URLs, e.g. `studio.edx.org/media/scorm/.......`.
This setting is set to `DENY` on edx-platform cms and lms via the setting `X_FRAME_OPTIONS`.

This PR extends django's XFrameOptions middleware to override the cms or lms django setting via an override setting called `X_FRAME_OPTIONS_OVERRIDES = [[regex, header_value]]` where regex is a regex that matches urls to override and header_value is one of `DENY`, `SAMEORIGIN`, `ALLOW`.

**How to test:**

Since this is a library, the automatic tests should suffice.

If you still want to test it with edx-platform to make sure it doesn't break anything, let me know and I can walk you through it.

### DIFF
--- a/edx_django_utils/security/clickjacking/README.rst
+++ b/edx_django_utils/security/clickjacking/README.rst
@@ -1,0 +1,28 @@
+Security Utils
+##############
+
+Common security utilities.
+
+Deprecated - Clickjacking middleware
+************************************
+
+Consider this deprecated as new projects should be using Content Security Policy headers instead.
+However, this won't be removed until no openedx project depends on it anymore, specifically, edx-platform.
+Please do not add this middleware to a project unless you have to use the ``X-Frame-Options`` header,
+and do not add it outside the openedx github organization, as it may be removed in the future.
+
+Clickjacking middleware
+***********************
+
+The preferred way to prevent clickjacking is to use the Content Security Policy headers.
+In some legacy code where those headers are not used, it is required to instead use the older
+header ``X-Frame-Options`` set either to ``DENY`` or to ``SAMEORIGIN``.
+In any case, this middleware allows you both to set the ``X-Frame-Options`` header to any recognized value -
+``DENY``, ``SAMEORIGIN``, ``ALLOW`` per django setting - but defaults to ``DENY``.
+It also allows you to override the header for specific urls defined via regex.
+
+- Add the middleware ``'edx_django_utils.security.clickjacking.middleware.EdxXFrameOptionsMiddleware'`` near the end of your ``MIDDLEWARE`` list.
+- This will add an `X-Frame-Options` header to all responses.
+- Add ``X_FRAME_OPTIONS = value`` to your django settings file with "value" being ``DENY``, ``SAMEORIGIN``, or ``ALLOW``.
+- Optionally, add ``X_FRAME_OPTIONS_OVERRIDES = [[regex, value]]`` where ``[[regex, value]]`` is a list of
+  pairs consisting of a regex that matches urls to override and a value that's one of ``DENY``, ``SAMEORIGIN``, and ``ALLOW``.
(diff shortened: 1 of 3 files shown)


edx_django_utils/security/clickjacking/tests/test_middleware.py
```diff
@@ -0,0 +1,119 @@
...
+        # settings.X_FRAME_OPTIONS_OVERRIDES = [['regex', 'SAMEORIGIN']]
+        # search.return_value = False
+        # middleware.process_response(request, response)
+        # assert response.headers['X-Frame-Options'] == 'SAMEORIGIN'
+        # validate_header.assert_called_with('SAMEORIGIN')
+        # search.assert_called_with('regex')
+        # assert response.headers['X-Frame-Options'] == 'SAMEORIGIN'
+        # validate_header.assert_called_with('SAMEORIGIN')
+        # search.assert_called_with('regex')
+        # assert response.headers['X-Frame-Options'] == 'SAMEORIGIN'
```
robrap, line 119: Code that should be deleted.
```suggestion
```

edx_django_utils/security/clickjacking/README.rst
```diff
@@ -0,0 +1,28 @@
...
+
+Clickjacking middleware
+***********************
+
+The preferred way to prevent clickjacking is to use the Content Security Policy headers.
+In some legacy code where those headers are not used, it is required to instead use the older
+header ``X-Frame-Options`` set either to ``DENY`` or to ``SAMEORIGIN``.
+In any case, this middleware allows you both to set the ``X-Frame-Options`` header to any recognized value -
+``DENY``, ``SAMEORIGIN``, ``ALLOW`` per django setting - but defaults to ``DENY``.
+It also allows you to override the header for specific urls defined via regex.
```
robrap, line 22: I'm not clear. You are adding a deprecated middleware with a new feature? So this regex feature will not be needed in some future time?

robrap (commented):
I did not do a detailed review. I'm just leaving a high-level question. Thanks.

edx_django_utils/security/clickjacking/tests/test_middleware.py
jesperhodge: Oops sorry. Yeah I was just going through cleanup, fixed that and going through linter errors. I should have marked this as draft perhaps

edx_django_utils/security/clickjacking/README.rst
jesperhodge: Okay let me try to explain my intention. Probably there's a better way to do this.

From a security perspective, we prefer CSP over X-Frame-Options for new or flexible projects. That means that it's okay to use X-Frame-Options but only if the project already has them and it's more work than it's worth to change them. That would be way outside the scope of my task for edx-platform - edx-platform still using the `X-Frame-Options` header.

The new middleware is important because we currently can't support SCORM xblock without setting the header to `SAMEORIGIN`, but I did not want to blindly do this for all of edx-platform. So I wrote this code to allow us to change the header for specific url patterns.

I wanted to mark this as deprecated because on the one hand we need it and it's not worth the effort to change the whole security headers thing on edx-platform right now, but I want to warn people off from using this middleware together with the outdated header in new projects.

Do you have a better suggestion for this?

robrap: I may have more thoughts, but a quick question is whether this is for edx-platform only, and if it should live there? That would reduce scope of introducing deprecated-from-the-start code.

timmc-edx:
This PR can be closed now, right?

PR #435, opened 2024-07-29
feat: Create manual spans for monitoring backends.

dianakhuang:
**Ticket:**
https://github.com/edx/edx-arch-experiments/issues/658

**Description:**

This adds a new context manager for creating manual spans across our various backends. It adds in an interface for passing through arbitrary backend data because Datadog, New Relic, and OTEL all have different interfaces for creating these spans.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,12 @@ Change Log
 
 .. There should always be an "Unreleased" section for changes pending release.
 
+[5.15.0] - 2024-07-29
+---------------------
+Added
+~~~~~
+* Added Datadog implementation of ``function_trace`` and allowed implementation to be configurable.
+
 [5.14.2] - 2024-05-31
 ---------------------
 Fixed
--- a/edx_django_utils/monitoring/internal/backends.py
+++ b/edx_django_utils/monitoring/internal/backends.py
@@ -50,6 +50,18 @@ class TelemetryBackend(ABC):
         Record the exception that is currently being handled.
         """
 
+    @abstractmethod
+    def create_span(self, name):
+        """
+        Start a tracing span with the given name, returning a context manager instance.
+
+        The caller must use the return value in a `with` statement or similar so that the
+        span is guaranteed to be closed appropriately.
+
+        Implementations should create a new child span parented to the current span,
+        or create a new root span if not currently in a span.
+        """
+
 
 class NewRelicBackend(TelemetryBackend):
     """
@@ -77,6 +89,13 @@ class NewRelicBackend(TelemetryBackend):
         # https://docs.newrelic.com/docs/apm/agents/python-agent/python-agent-api/recordexception-python-agent-api/
         newrelic.agent.record_exception()
 
+    def create_span(self, name):
+        if newrelic.version_info[0] >= 5:
+            return newrelic.agent.FunctionTrace(name)
+        else:
+            nr_transaction = newrelic.agent.current_transaction()
+            return newrelic.agent.FunctionTrace(nr_transaction, name)
+
 
 class OpenTelemetryBackend(TelemetryBackend):
     """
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -11,6 +11,12 @@ Change Log
 
 .. There should always be an "Unreleased" section for changes pending release.
 
+[5.15.0] - 2024-07-29
+---------------------
+Added
+~~~~~
+* Added ``start_trace`` to ``monitoring.utils`` so that manual spans in our monitoring backends can be created.
```
timmc-edx, line 18: We should probably use the public location of the addition in the changelog:

```suggestion
* Added ``start_trace`` to ``edx_django_utils.monitoring`` so that manual spans in our monitoring backends can be created.
```

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -92,6 +94,18 @@ def record_exception():
...
+@contextmanager
+def start_trace(name, params=None):
+    """
+    A context manager that starts spans for all configured backends.
+    """
+
+    # ExitStack handles the underlying context managers.
+    with ExitStack() as stack:
+        for backend in configured_backends():
+            yield stack.enter_context(backend.trace(name, params=params))
```
timmc-edx, line 106: Still reading up on ExitStack and how this interacts with other context managers, but I think we want a single yield after the loop:

```suggestion
            stack.enter_context(backend.trace(name, params=params))
        yield
```

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -89,6 +103,7 @@ def __init__(self):
         # If import fails, the backend won't be used.
         from opentelemetry import trace
         self.otel_trace = trace
+        self.otel_tracer = trace.get_tracer(__name__)
```
timmc-edx, line 106: This line can be dropped.

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -119,6 +138,9 @@ def record_exception(self):
         if span := self.dd_tracer.current_span():
             span.set_traceback()
 
+    def trace(self, name):
+        return self.dd_tracer.trace(name)
+
```
timmc-edx, line 143: We should add Datadog to the compatibility table in the monitoring README file.

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -50,6 +50,13 @@ def record_exception(self):
         Record the exception that is currently being handled.
         """
 
+    @abstractmethod
+    def trace(self, name):
+        """
+        Create a manual trace with the given name. The underlying implementation is generally
+        assumed to be a contextmanager, and the level above calls it as such.
```
timmc-edx, line 57: This might be clearer, see what you think:

```suggestion
        Start a tracing span with the given name, returning a context manager instance.
        
        The caller must use the return value in a `with` statement or similar so that the
        span is guaranteed to be closed appropriately.
        
        Implementations should create a new child span parented to the current span,
        or create a new root span if not currently in a span.
```

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -50,6 +50,13 @@ def record_exception(self):
         Record the exception that is currently being handled.
         """
 
+    @abstractmethod
+    def trace(self, name):
```
timmc-edx, line 54: Nitpicking on the naming... maybe something like this?

```suggestion
    def new_span(self, name):
```

I don't love this suggestion, but "trace" doesn't feel great either. I'd like to convey "we're opening/activating a span" but in particular I want to avoid any implication that we're creating a brand-new _trace_ (with a new root).

I never know what to name context managers anyhow. -.-

PR #451, opened 2024-09-26
feat: Add root span tagging for exceptions.

dianakhuang:
Datadog has issues with tagging the root span
with error information. This change adds the functionality
to tag the root span on exceptions.

This also renames the CachedCustomMonitoringMiddleware into
MonitoringSupportMiddleware so that it can be a central place
for most monitoring middleware changes instead of adding a new
one every time. At some point, CachedCustomMonitoringMiddleware
will be removed.

https://github.com/edx/edx-arch-experiments/issues/647

### DIFF
--- a/edx_django_utils/monitoring/internal/backends.py
+++ b/edx_django_utils/monitoring/internal/backends.py
@@ -62,6 +62,14 @@ class TelemetryBackend(ABC):
         or create a new root span if not currently in a span.
         """
 
+    @abstractmethod
+    def tag_root_span_with_error(self, exception):
+        """
+        Tags the root span with the given exception. This is primarily useful for
+        Datadog as New Relic handles this behavior correctly. Unclear if this is also needs to
+        be implemented for OTEL.
+        """
+
 
 class NewRelicBackend(TelemetryBackend):
     """
@@ -96,6 +104,10 @@ class NewRelicBackend(TelemetryBackend):
             nr_transaction = newrelic.agent.current_transaction()
             return newrelic.agent.FunctionTrace(nr_transaction, name)
 
+    def tag_root_span_with_error(self, exception):
+        # Does not need to be implemented for NewRelic, because it is handled automatically.
+        pass
+
 
 class OpenTelemetryBackend(TelemetryBackend):
     """
@@ -121,6 +133,10 @@ class OpenTelemetryBackend(TelemetryBackend):
         # Currently, this is not implemented.
         pass
 
+    def tag_root_span_with_error(self, exception):
+        # Currently, this is not implemented for OTel
+        pass
+
 
 class DatadogBackend(TelemetryBackend):
     """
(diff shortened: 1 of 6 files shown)


edx_django_utils/monitoring/internal/backends.py
```diff
@@ -62,6 +62,14 @@ def create_span(self, name):
         or create a new root span if not currently in a span.
         """
 
+    @abstractmethod
+    def tag_root_span(self, exception):
```
robrap, line 66: Maybe rename to `tag_root_span_with_error`, or something like that? This reads like you will be adding general span tags to the root span.

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -70,8 +70,10 @@ def record_python_version():
         _set_custom_attribute('python_version', platform.python_version())
 
 
-class CachedCustomMonitoringMiddleware(MiddlewareMixin):
+class MonitoringSupportMiddleware(MiddlewareMixin):
     """
+    Middleware to support monitoring, including cached custom attributes
+
     Middleware batch reports cached custom attributes at the end of a request.
```
robrap, line 77: Possible update:
```suggestion
    Middleware to support monitoring.

    1. Middleware batch reports cached custom attributes at the end of a request.
    2. Middleware adds error span tags to the root span.
```

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -145,6 +154,16 @@ def process_exception(self, request, exception):    # pylint: disable=W0613
...
         self._batch_report()
+        self._tag_root_span(exception)
+
+
+class CachedCustomMonitoringMiddleware(MonitoringSupportMiddleware):
+    """
+    DEPRECATED: Use MonitoringSupportMiddleware instead.
+
+    This is the old name for the MonitoringSupportMiddleware. We are keeping it
+    around for backwards compatibility until it can be fully deprecated and removed.
```
robrap, line 165: Nit:
```suggestion
    around for backwards compatibility until it can be fully removed.
```

edx_django_utils/monitoring/internal/middleware.py
```diff
@@ -145,6 +154,16 @@ def process_exception(self, request, exception):    # pylint: disable=W0613
         Django middleware handler to process an exception
         """
         self._batch_report()
+        self._tag_root_span(exception)
```
robrap, line 157: Can we add a test, especially so this functionality doesn't get lost by mistake at some point?

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -145,6 +161,10 @@ def record_exception(self):
     def create_span(self, name):
         return self.dd_tracer.trace(name)
 
+    def tag_root_span_with_error(self, exception):
+        root_span = self.dd_tracer.current_root_span()
+        root_span.set_exc_info(type(exception), exception, exception.__traceback__)
```
robrap, line 166: Is it possible to mock this instead to get a deeper mock and to fix this codecov comment?

robrap (approved):
Looks great. Thanks.

PR #453, opened 2024-10-01
docs: Add more context to CSP ADR

timmc-edx:
**Merge checklist:**
- [x] Version bumped
- [x] Changelog record added
- [x] Commits are squashed


docs/decisions/0006-content-security-policy-middleware.rst
```diff
@@ -37,7 +41,7 @@ We will enable this middleware for all Django-based IDAs.
 Consequences
 ************
 
-Django-based IDAs will for now be restricted to using static headers, and will not support ``strict-dynamic``. This is equivalent in configurability to having a CDN attach headers, and is good enough for a first pass, but will almost certainly need to be replaced with a more integrated system later as IDAs are adjusted to become amenable to ``strict-dynamic``.
+Django-based IDAs will for now be restricted to using static headers, and will not support ``strict-dynamic`` or view-specific variation. This is equivalent in configurability to having a CDN attach headers, and is good enough for a first pass, but will almost certainly need to be replaced with a more integrated system later as IDAs are adjusted to become amenable to ``strict-dynamic`` and we've already collected all the gains we can achieve with service-wide policies.
```
robrap, line 44: This looks fine, but I'm wondering what this means (vs the rejected alternative about arbitrary headers). Maybe a note about the implementation of this? Where and how are these static headers set?

Note: This is a question I have about the ADR in general, and not directly related to your changes in this PR.

robrap (approved):
I have a question that could lead to an improvement in this PR, or another PR, but isn't blocking.

docs/decisions/0006-content-security-policy-middleware.rst
timmc-edx: The differences between this and the arbitrary-headers alternative are quite minor. They would both entail static, service-wide headers that don't support strict-dynamic, view-specific values, etc. The main difference is just ease of configuration (breaking up the value across multiple lines).

robrap: Something like this, including an example of what each looks like, would make this more clear. That said - you can determine whether or not you'd like to make it more clear.

PR #457, opened 2024-10-15
feat: Add implementation for set_monitoring_transaction_name in DD.

dianakhuang:
This also refactors the existing NewRelic implementation to use the backends framework.

Also added comments about not porting ignore_transactions. This way of ignoring spans is not supported by Datadog.

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


edx_django_utils/monitoring/internal/backends.py
```diff
@@ -70,6 +70,12 @@ def tag_root_span_with_error(self, exception):
         be implemented for OTEL.
         """
 
+    @abstractmethod
+    def set_current_span_name(self, name, group=None, priority=None):
```
timmc-edx, line 74: For NR, the "transaction" most closely maps to what Datadog calls the local root span. I don't think we want to modify the current span in either case.

```suggestion
    def set_local_root_span_name(self, name, group=None, priority=None):
```

dianakhuang: renamed it.

edx_django_utils/monitoring/README.rst
```diff
@@ -32,7 +32,11 @@ Feature support matrix for built-in telemetry backends:
      - ✅
      - ❌
      - ✅
-   * - Retrieve and manipulate spans (``get_current_transaction``, ``ignore_transaction``, ``set_monitoring_transaction_name``)
+   * - Set span name (``set_monitoring_transaction_name``)
```
timmc-edx, line 35:
```suggestion
   * - Set local root span name (``set_monitoring_transaction_name``)
```

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -165,6 +178,11 @@ def tag_root_span_with_error(self, exception):
         root_span = self.dd_tracer.current_root_span()
         root_span.set_exc_info(type(exception), exception, exception.__traceback__)
 
+    def set_local_root_span_name(self, name, group=None, priority=None):
+        # For Datadog, this updates the 'operation_name' to the given name.
+        current_span = self.dd_tracer.current_root_span()
+        current_span.name = name
```
timmc-edx, line 184:
```suggestion
        local_root_span = self.dd_tracer.current_root_span()
        local_root_span.name = name
```

edx_django_utils/monitoring/internal/backends.py
```diff
@@ -165,6 +178,11 @@ def tag_root_span_with_error(self, exception):
         root_span = self.dd_tracer.current_root_span()
         root_span.set_exc_info(type(exception), exception, exception.__traceback__)
 
+    def set_local_root_span_name(self, name, group=None, priority=None):
+        # For Datadog, this updates the 'operation_name' to the given name.
```
timmc-edx, line 182: Does it not also update the resource name? I think that's the main thing we want to change, here. (NR does not do a good job of documenting what the "group" is, but I think it's actually mostly similar to DD's operation_name.)

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -112,6 +112,16 @@ def function_trace(function_name):
         yield
 
 
+def set_monitoring_transaction_name(name, group=None, priority=None):
+    """
+    Sets the transaction name for monitoring.
+
+    This is not cached.
```
timmc-edx, line 119:
```suggestion
    Sets the name, group, and priority for the current root span.

    Current root span refers to the most ancestral span in the current process, rather than
    to the trace root, which may be outside of the process.

    Group and priority may not be supported by all backends.
```
(I suspect the "not cached" note is an artifact that is no longer meaningful. At least, I have no idea what it means here...)

timmc-edx (commented):
Some suggestions on docs, and a question or two.

edx_django_utils/monitoring/internal/backends.py
dianakhuang: In my testing, it didn't, and I couldn't find anything that would let me set it.

Issue #419, opened 2024-05-30
Add Datadog support for span-related monitoring functions


timmc-edx:
[Consider breaking out each method as its own ticket when this is to be picked up.] 

Acceptance criteria:

- [ ] Complete all of the following for each method:
  - [x] `function_trace`: https://github.com/openedx/edx-django-utils/pull/435
    - [x] Move implementation into `NewRelicBackend`.
    - [x] Implement in `DatadogBackend`.
    - [x] Update function to call all of the telemetry backends.
    - [x] `edx_django_utils/monitoring/README.rst` updated with latest compatibility info.
    - [x] Inform owners of calling code, especially for backward-incompatible changes in Datadog.
  - [x] `set_monitoring_transaction_name`
    - [x] Move implementation into `NewRelicBackend`.
    - [x] Implement in `DatadogBackend`.
    - [x] Update function to call all of the telemetry backends.
    - [x] `edx_django_utils/monitoring/README.rst` updated with latest compatibility info.
    - [x] Inform owners of calling code, especially for backward-incompatible changes in Datadog (see notes).
  - [x] `ignore_transaction`
    - [x] Move implementation into `NewRelicBackend`.
    - [x] Implement in `DatadogBackend`.
    - [x] Update function to call all of the telemetry backends.
    - [x] `edx_django_utils/monitoring/README.rst` updated with latest compatibility info.
    - [x] Inform owners of calling code, especially for backward-incompatible changes in Datadog.
  - [ ] Handle `get_current_transaction` appropriately (see notes).
  - [ ] Handle `background_task` appropriately (see notes).

Implementation notes:

- Optionally, choose new names that are less specific to NR, and deprecate the old names (leaving stubs in their place that emit a warning and call the new function).

Other notes:

- Some repos currently call these directly on `newrelic.agent` rather than using edx-django-utils's indirection. https://github.com/edx/edx-arch-experiments/issues/621 covers resolving this.
- It appears that `get_current_transaction` is currently only used internally, despite being exposed, so it can probably be skipped for now.
- Relatedly, `background_task` appears to be completely unused. (Should this be retired?)
- 2U did not end up using OpenTelemetry, so we can't commit to adding support to OpenTelemetryBackend at the moment. These new methods can be implemented as no-ops in that backend.
[...]

PR #458, opened 2024-10-16
feat: Remove unusued background_task transaction.

dianakhuang:
Also add comments explaining get_current_transaction is internal-only and will not be exposed via API.

https://github.com/openedx/edx-django-utils/issues/419
**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [x] Version bumped
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


edx_django_utils/monitoring/internal/transactions.py
```diff
@@ -53,7 +53,9 @@ def name(self):
 
 def get_current_transaction():
     """
-    Returns the current transaction.
+    Returns the current transaction. This is only used internally and won't
```
robrap, line 56: Can this need to be removed from`edx_django_utils/monitoring/__init__.py` and docs?

CHANGELOG.rst
```diff
@@ -11,6 +11,13 @@ Change Log
 
 .. There should always be an "Unreleased" section for changes pending release.
 
+Unreleased
+----------
+Changed
+~~~~~~~
+* Remove unused ``background_task`` monitoring function.
+* Remove ``get_current_transaction`` from the public API.
```
robrap, line 19: [question]: Do we want to make this a 7.0.0 breaking change because that is how the contract phase is typically handled (i.e. we get things so that they are unused, and then we remove them)? It seems right, even though it is strange. Either way, I think we need to decide and do it as part of this PR, rather than leaving this to some future person to decide. Thoughts?

```suggestion
Removed
~~~~~~~
* Remove unused ``background_task`` monitoring function.
* Remove ``get_current_transaction`` (used internally only) from the public API.
```

dianakhuang: Yeah, I thought these changes would be more minor than they ended up being. I'll bump the version number so we release it as 7.0.0.

robrap (approved):

PR #478, opened 2025-02-03
Add support for kwargs in create_span method

hamza-56:
### Add support for kwargs in create_span method

Added support for passing **kwargs to the create_span method in the TelemetryBackend abstract class. This allows backends like **DatadogBackend** to accept additional arguments (e.g., resource) when creating spans.

### DIFF
--- a/edx_django_utils/monitoring/internal/backends.py
+++ b/edx_django_utils/monitoring/internal/backends.py
@@ -51,7 +51,7 @@ class TelemetryBackend(ABC):
         """
 
     @abstractmethod
-    def create_span(self, name):
+    def create_span(self, name, **kwargs):
         """
         Start a tracing span with the given name, returning a context manager instance.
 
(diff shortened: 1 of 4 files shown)


edx_django_utils/monitoring/internal/backends.py
```diff
@@ -51,7 +51,7 @@ def record_exception(self):
         """
 
     @abstractmethod
-    def create_span(self, name):
+    def create_span(self, name, **kwargs):
```
robrap, line 54: I was imaging that the API would be well defined, and not completely open. Note that someone using this API needs to write a single line of open source code (e.g. `create_span(name, xxx)` etc.) that should do the right thing no matter what backend is configured.  So, I imagining a well defined second argument, with documentation about how it is treated for each implementing backend.

hamza-56:
Closing this PR in favor of using `set_monitoring_transaction_name` along with `function_trace`

PR #482, opened 2025-02-18
Add monitor_django_management_command

hamza-56:
### Add monitor_django_management_command

monitor_django_management_command creates a monitoring span using `function_trace`, 
allowing the execution of the command to be tracked explicitly in monitoring tools.

It also sets the transaction name to the given `name` using `set_monitoring_transaction_name`.


edx_django_utils/monitoring/internal/utils.py
```diff
@@ -19,9 +19,23 @@
...
 from contextlib import ExitStack, contextmanager
 
+from django.conf import settings
+
 from .backends import configured_backends
 from .middleware import CachedCustomMonitoringMiddleware
 
+DJANGO_MANAGEMENT_MONITORING_OPERATION_NAME = getattr(
+    settings,
+    'DJANGO_MANAGEMENT_MONITORING_OPERATION_NAME',
```
robrap, line 29: What do you think of the following?
```suggestion
    'DJANGO_MANAGEMENT_MONITORING_FUNCTION_TRACE_NAME',
```

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -143,3 +157,22 @@ def noop_decorator(func):
...
+
+@contextmanager
+def monitor_django_management_command(name):
+    """
+    A context manager for monitoring Django management commands.
+
+    This function creates a monitoring span using `function_trace`, allowing
+    the execution of the command to be tracked explicitly in monitoring tools.
+    It also sets the transaction name to the given `name` using
+    `set_monitoring_transaction_name`.
```
robrap, line 170: I don't want the second point to get lost. What about?: 
```suggestion
    - This function creates a monitoring span using `function_trace`, allowing
      the execution of the command to be tracked explicitly in monitoring tools.
    - It also sets the transaction name to the given `name` using
      `set_monitoring_transaction_name`.
```

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -143,3 +157,22 @@ def noop_decorator(func):
         return newrelic.agent.background_task(*args, **kwargs)
     else:
         return noop_decorator
+
+
+@contextmanager
+def monitor_django_management_command(name):
```
robrap, line 163: Please document in https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/monitoring/README.rst.

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -143,3 +157,22 @@ def noop_decorator(func):
...
+    """
+    A context manager for monitoring Django management commands.
+
+    This function creates a monitoring span using `function_trace`, allowing
+    the execution of the command to be tracked explicitly in monitoring tools.
+    It also sets the transaction name to the given `name` using
```
robrap, line 169: Is `name` the name of the management command? Maybe something like: "It also sets the transaction name to the name of the management command, provided in the `name` argument.

edx_django_utils/monitoring/internal/utils.py
```diff
@@ -19,9 +19,23 @@
 """
 from contextlib import ExitStack, contextmanager
 
+from django.conf import settings
+
 from .backends import configured_backends
 from .middleware import CachedCustomMonitoringMiddleware
 
+DJANGO_MANAGEMENT_MONITORING_OPERATION_NAME = getattr(
```
robrap, line 27: Can you annotate the settings. Follow an example like: https://github.com/openedx/edx-platform/blob/45f44c3965498fef80dcde676f2969324820a590/openedx/core/djangoapps/oauth_dispatch/jwt.py#L125-L130

robrap (commented):
Thanks. Is this something you've tested using (2U private link) https://2u-internal.atlassian.net/wiki/spaces/ENG/pages/1173618788/Running+Datadog+in+devstack?

hamza-56:
> Thanks. Is this something you've tested using (2U private link) https://2u-internal.atlassian.net/wiki/spaces/ENG/pages/1173618788/Running+Datadog+in+devstack?

Yes, I have

[
Screenshot 2025-02-25 at 2 58 25 AM
](url)

robrap (approved):
Thanks!

edx_django_utils/monitoring/README.rst
```diff
@@ -126,3 +126,21 @@ Monitoring Memory Usage
...
+Here's an example of how to integrate it into your ``manage.py``:
+
+.. code-block:: python
+
+    if __name__ == "__main__":
+        ...
+        from django.core.management import execute_from_command_line
+        from edx_django_utils.monitoring import monitor_django_management_command
+
+        with monitor_django_management_command(sys.argv[1]):
```
regisb, line 145: Is it expected that this context manager will be added to manage.py in some public Django apps? I'm concerned about the potential overhead, in particular in terms of import time. For instance, edx-platform is notorious for its very slow manage.py. If we decorate every manage.py with this context manager, we will be pulling all the imports from edx_django_utils/monitoring/__init__.py in every command.

regisb: Are you considering adding this context manager to some public Django apps?

robrap: @regisb: Are these times ok with you? If not, we can update the example code with a comment that checks `ENABLE_CUSTOM_MANAGEMENT_COMMAND_MONITORING` (without referencing edx-django-utils), and decides whether to import or not.

regisb: For reference, this is the output of a simple command in edx-toggles on my computer:

```$ time ./manage.py shell -c "print()"
real    0m0,394s
user    0m0,335s
sys     0m0,059s```

So that 0.15s overhead is actually ~35% of the total time. In absolute terms that's not a big deal, but this added lag is really a code smell: it's an indication that we are adding a whole lot of complexity on the path of every single manage.py command. And I'm not quite convinced about the usefulness of that feature, since I'm guessing that 99% of Open edX platforms out there don't define OPENEDX_TELEMETRY.

This is why I was asking whether you intend to add this piece of code to public Django apps. If this feature is limited to 2U-only apps, then I don't have any issue. But if the goal is to add this context manager to multiple apps, then my opinion would be that this is too much bloat.

Comment (robrap) [reply edx_django_utils/monitoring/README.rst:145]: I think the intention is to add it to Open edX services, so I think we should just code it in a way that skips the import if this isn't enabled.

Comment (hamza-56) [reply edx_django_utils/monitoring/README.rst:145]: Our intention is to add this into Open edX services. For now, we will integrate it only with course discovery.
Other teams can integrate this if they want monitoring support for their management commands.

Comment (regisb) [reply edx_django_utils/monitoring/README.rst:145]: I think it's a bad pattern to import large libraries in manage.py. If I wanted to monitor manage.py commands, I think I would avoid modifying the manage.py code, and instead use a CLI similar to `myinstrumenter ./manage.py <args>`.

Plus, there's nothing special about the `monitor_django_management_command`: this function is just a thin wrapper on top of `function_trace`. So I think that neither the changes to manage.py nor the added function to edx_django_utils are really necessary. The added bloat, in the form of two new cross-app Django settings and custom code on the path of every call to manage.py are just not worth it.

But I'm just one guy with an opinion here. Feel free to discuss this with other people and ignore my comment.

Comment (hamza-56) [reply edx_django_utils/monitoring/README.rst:145]: I understand the concern about adding bloat and modifying `manage.py`, but we [already use a CLI](https://github.com/edx/helm-charts/blob/master/django-ida/templates/cronjobs.yaml#L37) (ddtrace-run). Even using this CLI, **Datadog** does not automatically instrument Django management commands, unlike **New Relic**, which already offers this functionality.

We raised a **Datadog** support ticket ([#1990940](https://help.datadoghq.com/hc/en-us/requests/1990940?page=1)) to explore options, as we want to avoid adding custom instrumentation for every command.

Then we opted to use `function_trace`. However, the problem we're facing is that the `function_trace` command takes a single argument that is used for both the `operation` and `resource` names.

This limitation means that management commands do not get a distinct, consistent `operation_name` (e.g., `django.command`), which is necessary for proper observability. 

To address this, we implemented `monitor_django_management_command()`. This packages all the boilerplate logic, ensuring minimal modifications to `manage.py` while achieving the desired instrumentation.

Comment (hamza-56) [reply edx_django_utils/monitoring/README.rst]: I have updated the example code to only import if `ENABLE_CUSTOM_MANAGEMENT_COMMAND_MONITORING` is enabled

Comment (robrap) [reply edx_django_utils/monitoring/README.rst]: @regisb: I don't know of a way to inject the code we need from outside `manage.py`, but we can explore that a little more to see if we missed anything.

If we **do** need to add this monitoring capability to services in general, this new thin wrapper is also wrapping other calls, like `function_trace`, that are a part of our larger abstraction layer to various monitoring tools. I think it makes sense to document in one place and use the abstraction layer. For those that don't require monitoring, this will be disabled by default.

File: edx_django_utils/monitoring/README.rst [hunk 7]
```diff
@@ -126,3 +126,28 @@ Monitoring Memory Usage
...
+.. code-block:: python
+
+    if __name__ == "__main__":
+        ...
+        from django.conf import settings
+        from django.core.management import execute_from_command_line
+
+        monitoring_enabled = getattr(settings, "ENABLE_CUSTOM_MANAGEMENT_COMMAND_MONITORING", False)
+
+        if monitoring_enabled and len(sys.argv) > 1:
```
robrap, line 147: What is a scenario where `len(sys.argv) > 1`, and would you really want to drop monitoring altogether, or would you want to just pass a different (static) value to `monitor_django_management_command`?

hamza-56: When `manage.py` is run without any command or extra arguments: `len(sys.argv) <= 1`.
When `manage.py` is run with a command (e.g., `python manage.py migrate`): `len(sys.argv) > 1`.

robrap: Does `execute_from_command_line` just report an error, and does this conditional simply allow the same error to continue to report? If so, I get it and that sounds fine.

hamza-56: When `manage.py` is run without a command (e.g., `python manage.py`), Django's `execute_from_command_line(sys.argv)` handles this gracefully by printing usage instructions instead of executing a command. 

The conditional simply prevents `sys.argv[1]` from being accessed when no command is provided, avoiding a crash.

Screenshot 2025-03-01 at 6 16 28 AM

edx_django_utils/monitoring/README.rst
robrap: @regisb:
1. The example code has been updated to avoid imports when not needed.
2. We did not find an alternative to this code for adding observability.
3. My opinion is that duplicating all of this in each service would be worse than the current solution.

Is this satisfying enough, or do you have an improved proposal?

regisb: I'd like to reiterate that I'm just stating my opinion here, and I'm not blocking this PR in any manner. You should feel free to merge and update other repos as you see fit.

My concern is chiefly about edX adding features to the Open edX codebase that are useless for most platforms -- features which thus qualify as bloat for these platforms.

I see at least three different ways in which edX could implement this feature without affecting other platforms:

1. Implement a `edx-manage.py` command which would be pip-installed in your production platforms, but would not reside in the upstream repo. In this script you would call `execute_from_command_line` and any other piece of code that edX considers necessary.
2. Using an edX-specific Django app, monkey-patch all `Command` objects at runtime to enable monitoring for these commands. This app would be added to your `INSTALLED_APPS` in production.
3. Using the Open edX hooks framework, implement a new [filter](https://github.com/openedx/openedx-filters) that would pass the command to be run. edX would then be able to add a callback to this filter that would decorate the corresponding command with the right tracing functions.

Imagine you were running [saleor](https://github.com/saleor/saleor), which is a popular Django project for e-commerce. Saleor has its own [manage.py](https://github.com/saleor/saleor/blob/main/manage.py), to which you would like to add tracing. Surely you wouldn't open a PR upstream to add tracing there, right? And thus you would implement one of the options that I outlined above. I argue that we are in a similar situation here with Open edX.

robrap: @regisb: I agree on the sentiment and really appreciate you making some of the options more explicit for the owning team. I can discuss with @hamza-56.

UsamaSadiq:
@hamza-56 @robrap do we have any update on this PR?

robrap:
Thanks for checking in @UsamaSadiq.

I'm going to close this PR:
- The solution from this PR was added to course-discovery specifically in https://github.com/openedx/course-discovery/pull/4594, rather than adding it here as a utility.
- A 2U backlog ticket https://github.com/edx/edx-arch-experiments/issues/958 was added to explore some of the other proposals made by Régis on this PR.

PR #492, opened 2025-03-28
chore: adding django 5.1 in tox and github actions.

awais786:
Adding django 5.1 in tox and GH actions.


UsamaSadiq:
`Django 5.2` support has been added in the PR https://github.com/openedx/edx-django-utils/pull/496 so this change is not needed anymore.

Issue #469, opened 2024-12-10
[DEPR]: Code Owner Monitoring


robrap:
### Proposal Date

2024-12-11

### Target Ticket Acceptance Date

2024-12-23

### Earliest Open edX Named Release Without This Functionality

Teak - 2025-04

### Rationale

The [code_owner monitoring code](https://github.com/openedx/edx-django-utils/tree/master/edx_django_utils/monitoring/internal/code_owner) in was introduced to help 2U monitor the large monolith (LMS), and attribute different parts of the app to different teams. This DEPR assumes that this is 2U-specific code, and that no other orgs or teams are using this functionality for monitoring purposes.

Slightly related, the New Relic search script was likely only used by 2U. This could be removed if no one is using it.

### Removal

What would be removed is https://github.com/openedx/edx-django-utils/tree/master/edx_django_utils/monitoring/internal/code_owner, and all associated code, config, and documentation.

Slightly related, we could also remove [the new_relic_search.py script](https://github.com/openedx/edx-django-utils/blob/18cf72da7503eb283447f631d54292e5111b56b5/edx_django_utils/monitoring/scripts/new_relic_search.py) and supporting documentation. There would be no replacement, other than simply making a copy of the script as-needed.

### Replacement

- It is possible that no one in the community needs a replacement.
- For 2U, the functionality was replaced by a plugin that adds instrumentation in two ways:
  1. For Django requests, instrumentation was added to the plugin using new signals that were added to edx-django-utils MonitoringSupportMiddleware, as documented in https://github.com/openedx/edx-django-utils/pull/467.
  2. For Celery tasks, 2U is using a Datadog span processor to add instrumentation.

For 2U, the span processor replaces the need for the custom decorator that was added to all Open edX celery tasks, and documented in the ADR: [0003-code-owner-for-celery-tasks](https://github.com/openedx/edx-django-utils/blob/65b138f55f13e1935f8e9155e6f443e020ec2ac1/edx_django_utils/monitoring/docs/decisions/0003-code-owner-for-celery-tasks.rst).

If we choose to remove this decorator, there is no real replacement if your monitoring tool does not have some way to hook into these tasks. New Relic had no such hook in the past.

### Deprecation

- It could simply be removed if no one is using it.
- The decorator for celery tasks is slightly more complicated, because it has references throughout the codebase. This might get marked as deprecated before removal.

### Additional Info

[...]

PR #547, opened 2026-08-10
chore: Remove deprecated code owner utilities and related tests

ktyagiapphelix2u:
### WARNING

The DEPR task list (https://github.com/openedx/edx-django-utils/issues/469) now clarifies what parts of this PR are unblocked and what is blocked. The blocked parts need to be removed for this to move forward.

### Summary
- Deleted the code owner utility functions and mappings from utils.py, which were no longer in use.
- Removed the New Relic search script that was dependent on the code owner utilities.
- Cleared out all associated test files for the code owner functionality, including middleware and utility tests.
- This cleanup is part of the effort to streamline the monitoring module and remove unused code.

### Ticket
https://2u-internal.atlassian.net/browse/BOMS-241


robrap (commented):
Some high-level feedback:
1. We need to determine whether we will add a deprecation warning before doing this removal.
2. We want to mark outdated ADRs as obsolete and add a new one to detail why, but ADRs shouldn't be removed.

ktyagiapphelix2u:
> * We need to determine whether we will add a deprecation warning before doing this removal.

@robrap I’m a bit unclear on this as well. From the original ticket https://2u-internal.atlassian.net/browse/BOMS-241, my understanding was that the deprecation warning should be added after the cleanup/removal work is completed.

Could you clarify what we need to determine before proceeding with the removal, and whether we’re expected to add the deprecation warning before the cleanup instead?

robrap:
> @robrap I’m a bit unclear on this as well. From the original ticket https://2u-internal.atlassian.net/browse/BOMS-241, my understanding was that the deprecation warning should be added after the cleanup/removal work is completed.

We can't have a deprecation warning for something that doesn't exist, so it needs to be added before the functionality is deleted. It is the first step, and it also read "_possibly_ after edx-platform clean-up", which I'm not sure why I added, but meant that we could begin removing uses in `edx-platform` (not definitions from `edx-django-utils`).

robrap:
[inform] The DEPR task list (https://github.com/openedx/edx-django-utils/issues/469) now clarifies what parts of this PR are unblocked and what is blocked. The blocked parts need to be removed for this to move forward.

PR #551, opened 2026-09-01
chore: Remove non-celery code owner monitoring code and docs

rahulkanneri-2u:
Removes the non-celery code owner monitoring code, docs, and the new_relic_search script, since this functionality is 2U-specific and has already been replaced by plugin-based instrumentation. The celery task decorator (set_code_owner_attribute) and its shared helpers are intentionally left in place, since it is still in active use and its removal is tracked separately.

**Description:**

Removes `CodeOwnerMonitoringMiddleware` and its supporting non-celery utilities, tests, and docs, plus the `new_relic_search.py` script and its docs. This is a trimmed-down replacement for #547, which mixed in premature removal of `utils.py` (the still-in-use celery decorator, `set_code_owner_attribute`). This PR leaves `utils.py`, `test_utils.py`, and the celery-related ADRs untouched.

**JIRA:**

openedx/edx-django-utils#469

**Dependencies:**

None.

**Merge deadline:**

None currently.

None.

**Testing instructions:**

None.

**Reviewers:**
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bumped
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

**Author concerns:**

migrations, etc.


naincy128 (approved):

robrap (commented):
I misspoke. We should not be making any breaking changes, so the middleware also should _not_ be removed. (Removal of the script could be considered a breaking change, but I am not counting that, so that is fine to stay.)

If we do anything, the most important change would be to add a [deprecation warning](https://github.com/search?q=org%3Aopenedx+warnings.warn&type=code) to any public api that would be removed (e.g. middleware, celery function, etc.). We would also update the changelog.

Note: we need to timebox or not do any changes here. Thanks.

rahulkanneri-2u:
Closing without merging. Per review (https://github.com/openedx/edx-django-utils/pull/551#pullrequestreview-5172299532).

`CodeOwnerMonitoringMiddleware` removal is a breaking change for downstream consumers and shouldn't ship without a
deprecation warning first. We're timeboxing this DEPR and not taking on the deprecation-warning work right now, so
deferring this cleanup rather than continuing to iterate. Tracked in openedx/edx-django-utils#469.

